PR #2, opened 2017-06-13
Use Maybe instead of throwing errors

vaibhavsagar:
I'm tossing around the idea of creating a drop-in replacement for `nix-serve` in Haskell and I thought I'd try to reduce the number of partial functions here.


vaibhavsagar:
How is `Unknown` better than `Nothing` in this case?

shlevy:
There's a semantic difference between "this is not a hash type" and "this is a hash type but I don't know enough about it to do anything useful".

vaibhavsagar:
As you wish :smile:

PR #3, opened 2018-04-27
ReadonlyStoreEffects: Add a referrers query.

shlevy:


hnix-store-core/src/System/Nix/Store.hs
```diff
@@ -41,6 +46,18 @@ type PathHashAlgo = Truncated SHA256 20
 -- | A path in a store.
 data Path = Path !(Digest PathHashAlgo) !PathName
 
+-- | Wrapper to defined a 'Hashable' instance for 'Digest'.
+newtype HashableDigest a = HashableDigest (Digest a)
+
+instance Hashable (HashableDigest a) where
+  hashWithSalt s (HashableDigest d) = unsafeDupablePerformIO $
+    B.withByteArray d $ \ptr -> hashPtrWithSalt ptr (B.length d) s
```
jwiegley: I think this use of `unsafeDupablePerformIO` is in keeping with what the FFI code does whenever you call a C function that hasn't been explicitly typed within the `IO` monad. I.e., since you're only reading data that won't be changed by other threads, it's still referentially transparent.

There's a decent discussion about this here: https://stackoverflow.com/questions/10529284/is-there-ever-a-good-reason-to-use-unsafeperformio?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

shlevy: Thanks!

PR #7, opened 2018-04-30
NAR parser/generator

imalsogreg:
NAR parser and generator. Moved `Path` types and functions into `Paths.hs` module.

### Todo
 - [x] cleanup
 - [x] test the generator manually
 - [x] create tests (parse . gen === id)
 - [x] finish the `binary` instance
 - [x] test encoding/decoding for large NARs (tried 10 and 12 Mb - found problems at 12)
   - can we stream in constant space? (moved to its own ticket)
   - what do we do with NARs beyond using them as intermediate form between download and unpack for computing store path hash? Provide API for doing these things (moved to its own ticket)
 - [x] factor IO out of `localUnpackNar` - use `NarEffects` monad instead
 - [x] Write an IO instance for `NarEffects`

### Testing Done

Our encoder and decoder, filesystem packer and filesystem unpacker are all checked against each other and against the output of `nix-store --dump`.

Some of our tests use generated files about 8MB in size, and a realistic directory (the `src` directory of this project) for head-to-head comparisons with `nix-store --dump`.

```
*NarFormat Main> main
tests/Driver.hs
  narEncoding
    parser-roundtrip
      roundtrips regular:    OK
      roundtrips regular 2:  OK
      roundtrips executable: OK
      roundtrips symlink:    OK
      roundtrips directory:  OK
    matches-nix-store fixture
      matches regular:       OK
      matches regular':      OK
      matches executable:    OK
      matches symlink:       OK
      matches directory:     OK
  nixStoreRegular:           OK (0.03s)
  nixStoreDirectory:         OK (0.03s)
  nixStoreDirectory':        OK (0.03s)
  nixStoreBigFile:           OK (0.15s)
  nixStoreBigDir:            OK (0.34s)
  narEncodingArbitrary:      OK (0.18s)
    +++ OK, passed 100 tests.
  packSelfSrcDir:            OK (0.03s)

All 17 tests passed (0.80s)
```

### DIFF
--- a/hnix-store-core/README.md
+++ b/hnix-store-core/README.md
@@ -7,3 +7,9 @@ See `StoreEffects` in [System.Nix.Store] for the available operations
 on the store.
 
 [System.Nix.Store]: ./src/System/Nix/Store.hs
+
+
+Tests
+======
+
+ - `ghcid --command "cabal repl test-suite:format-tests" --test="Main.main"`
--- a/hnix-store-core/hnix-store-core.cabal
+++ b/hnix-store-core/hnix-store-core.cabal
@@ -19,12 +19,51 @@ cabal-version:       >=1.10
 library
   exposed-modules:     Crypto.Hash.Truncated
                      , System.Nix.Nar
+                     , System.Nix.Path
                      , System.Nix.Store
-  build-depends:       base >=4.10 && <4.11,
-                       -- Drop foundation when we can drop cryptonite <0.25
-                       binary,
-                       bytestring, containers, cryptonite, memory, foundation, basement,
-                       text, regex-base, regex-tdfa-text,
-                       hashable, unordered-containers, bytestring
+  build-depends:       base >=4.10 && <4.11
+                     , basement
+                     , bytestring
+                     , binary
+                     , bytestring
+                     , containers
+                     , cryptonite
+                     , directory
+                     , filepath
+                      -- Drop foundation when we can drop cryptonite <0.25
+                     , foundation
+                     , hashable
+                     , memory
+                     , mtl
+                     , regex-base
+                     , regex-tdfa-text
+                     , text
+                     , unix
+                     , unordered-containers
   hs-source-dirs:      src
   default-language:    Haskell2010
+
+test-suite format-tests
+   ghc-options: -rtsopts -fprof-auto
+   type: exitcode-stdio-1.0
+   main-is: Driver.hs
+   other-modules:
+       NarFormat
+   hs-source-dirs:
+       tests 
+   build-depends:
+       hnix-store-core
+     , base
+     , base64-bytestring
+     , binary
+     , bytestring
+     , containers
+     , directory
+     , process
+     , tasty
+     , tasty-discover
+     , tasty-hspec
+     , tasty-hunit
+     , tasty-quickcheck
+     , text
+   default-language: Haskell2010
--- a/hnix-store-core/hnix-store-core.cabal
+++ b/hnix-store-core/hnix-store-core.cabal
@@ -17,10 +17,13 @@ extra-source-files:  ChangeLog.md, README.md
 cabal-version:       >=1.10
 
 library
-  exposed-modules:     Crypto.Hash.Truncated, System.Nix.Store
+  exposed-modules:     Crypto.Hash.Truncated
+                     , System.Nix.Nar
+                     , System.Nix.Store
   build-depends:       base >=4.10 && <4.11,
                        -- Drop foundation when we can drop cryptonite <0.25
-                       cryptonite, memory, foundation, basement,
+                       binary,
+                       bytestring, containers, cryptonite, memory, foundation, basement,
                        text, regex-base, regex-tdfa-text,
                        hashable, unordered-containers
   hs-source-dirs:      src
--- a/hnix-store-core/src/System/Nix/Nar.hs
+++ b/hnix-store-core/src/System/Nix/Nar.hs
@@ -1,181 +1,267 @@
-{-# LANGUAGE KindSignatures    #-}
-{-# LANGUAGE ScopedTypeVariables    #-}
-{-# LANGUAGE TupleSections    #-}
-{-# LANGUAGE TypeApplications  #-}
-{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections       #-}
+{-# LANGUAGE TypeApplications    #-}
 
 {-|
 Description : Allowed effects for interacting with Nar files.
 Maintainer  : Shea Levy <shea@shealevy.com>
 |-}
-module System.Nix.Nar where
+module System.Nix.Nar (
+    FileSystemObject(..)
+  , IsExecutable (..)
+  , Nar(..)
+  , getNar
+  , localPackNar
+  , localUnpackNar
+  , narEffectsIO
+  , putNar
+  ) where
 
-import           Control.Monad (replicateM, replicateM_)
-import           Data.Monoid ((<>))
 import           Control.Applicative
-import qualified Data.ByteString.Lazy.Char8 as BSL
-import qualified Data.Set as Set
-import qualified Data.Binary as B
-import qualified Data.Text as T
-import qualified Data.Text.Encoding as E
-import qualified Data.Binary.Put as B
-import qualified Data.Binary.Get as B
-import           Debug.Trace
+import           Control.Monad              (replicateM, replicateM_, (<=<))
+import qualified Data.Binary                as B
+import qualified Data.Binary.Get            as B
+import qualified Data.Binary.Put            as B
+import           Data.Bool                  (bool)
+import qualified Data.ByteString            as BS
+import qualified Data.ByteString.Char8      as BSC
+import qualified Data.ByteString.Lazy       as BSL
+import           Data.Foldable              (forM_)
+import qualified Data.Map                   as Map
+import           Data.Maybe                 (fromMaybe)
+import           Data.Monoid                ((<>))
+import qualified Data.Text                  as T
+import qualified Data.Text.Encoding         as E
+import           Data.Traversable           (forM)
+import           GHC.Int                    (Int64)
+import           System.Directory
+import           System.FilePath
+import           System.Posix.Files         (createSymbolicLink, fileSize, getFileStatus,
+                                             isDirectory, readSymbolicLink)
 
-import System.Nix.Path
+import           System.Nix.Path
 
-
-data NarEffects (m :: * -> *) = NarEffets {
-    readFile         :: FilePath -> m BSL.ByteString
-  , listDir          :: FilePath -> m [FileSystemObject]
-  , narFromFileBytes :: BSL.ByteString -> m Nar
-  , narFromDirectory :: FilePath -> m Nar
+data NarEffects (m :: * -> *) = NarEffects {
+    narReadFile   :: FilePath -> m BSL.ByteString
+  , narWriteFile  :: FilePath -> BSL.ByteString -> m ()
+  , narListDir    :: FilePath -> m [FilePath]
+  , narCreateDir  :: FilePath -> m ()
+  , narCreateLink :: FilePath -> FilePath -> m ()
+  , narGetPerms   :: FilePath -> m Permissions
+  , narSetPerms   :: FilePath -> Permissions ->  m ()
+  , narIsDir      :: FilePath -> m Bool
+  , narIsSymLink  :: FilePath -> m Bool
+  , narFileSize   :: FilePath -> m Int64
+  , narReadLink   :: FilePath -> m FilePath
 }
 
 
 -- Directly taken from Eelco thesis
 -- https://nixos.org/%7Eeelco/pubs/phd-thesis.pdf
 
--- TODO: Should we use rootedPath, validPath rather than FilePath?
 data Nar = Nar { narFile :: FileSystemObject }
-    deriving (Eq, Ord, Show)
+    deriving (Eq, Show)
 
+-- | A FileSystemObject (FSO) is an anonymous entity that can be NAR archived
 data FileSystemObject =
-    Regular IsExecutable BSL.ByteString
-  | Directory (Set.Set (PathName, FileSystemObject))
-  | SymLink BSL.ByteString
+    Regular IsExecutable Int64 BSL.ByteString
+    -- ^ Reguar file, with its executable state, size (bytes) and contents
+  | Directory (Map.Map FilePathPart FileSystemObject)
+    -- ^ Directory with mapping of filenames to sub-FSOs
+  | SymLink T.Text
+    -- ^ Symbolic link target
   deriving (Eq, Show)
 
--- TODO - is this right? How does thesis define ordering of FSOs?
-instance Ord FileSystemObject where
-    compare (Regular _ c1) (Regular _ c2) = compare c1 c2
-    compare (Regular _ _)  _              = GT
-    compare (Directory s1) (Directory s2) = compare s1 s2
-    compare (Directory _)  _              = GT
-    compare (SymLink l1) (SymLink l2)     = compare l1 l2
 
 data IsExecutable = NonExecutable | Executable
     deriving (Eq, Show)
 
--- data NarFile = NarFile
---     { narFileIsExecutable :: IsExecutable
---     , narFilePath         :: FilePath -- TODO: Correct type?
---     } deriving (Show)
 
-data DebugPut = PutAscii | PutBinary
+instance B.Binary Nar where
+  get = getNar
+  put = putNar
 
+------------------------------------------------------------------------------
+-- | Serialize Nar to lazy ByteString
 putNar :: Nar -> B.Put
-putNar = putNar' PutBinary
-
-putNar' :: DebugPut -> Nar -> B.Put
-putNar' dbg (Nar file) = header <>
-                         parens (putFile file)
+putNar (Nar file) = header <> parens (putFile file)
     where
 
-        str' = case dbg of
-            PutAscii -> strDebug
-            PutBinary -> str
+        header   = str "nix-archive-1"
 
-        header   = str' "nix-archive-1"
-        parens m = str' "(" <> m <> str ")"
-
-        putFile (Regular isExec contents) =
-               str' "type" <> str' "regular"
-            <> if isExec == Executable
-               then str' "executable" <> str' ""
-               else str' ""
-            <> str' "contents" <> str' contents
+        putFile (Regular isExec fSize contents) =
+               strs ["type", "regular"]
+            >> (if isExec == Executable
+               then strs ["executable", ""]
+               else return ())
+            >> putContents fSize contents
 
         putFile (SymLink target) =
-               str' "type" <> str' "symlink" <> str' "target" <> str' target
+               strs ["type", "symlink", "target", BSL.fromStrict $ E.encodeUtf8 target]
 
+        -- toList sorts the entries by FilePathPart before serializing
         putFile (Directory entries) =
-               str' "type" <> str' "directory"
-            <> foldMap putEntry entries
+               strs ["type", "directory"]
+            <> mapM_ putEntry (Map.toList entries)
 
-        putEntry (PathName name, fso) =
-            str' "entry" <>
-            parens (str' "name" <>
-                    str' (BSL.fromStrict $ E.encodeUtf8 name) <>
-                    str' "node" <>
-                    putFile fso)
+        putEntry (FilePathPart name, fso) = do
+            str "entry"
+            parens $ do
+              str "name"
+              str (BSL.fromStrict name)
+              str "node"
+              parens (putFile fso)
 
+        parens m = str "(" >> m >> str ")"
+
+        -- Do not use this for file contents
+        str :: BSL.ByteString -> B.Put
+        str t = let len = BSL.length t
+            in int len <> pad len t
+
+        putContents :: Int64 -> BSL.ByteString -> B.Put
+        putContents fSize bs = str "contents" <> int fSize <> (pad fSize bs)
+        -- putContents fSize bs = str "contents" <> int (BSL.length bs) <> (pad fSize bs)
+
+        int :: Integral a => a -> B.Put
+        int n = B.putInt64le $ fromIntegral n
+
+        pad :: Int64 -> BSL.ByteString -> B.Put
+        pad strSize bs = do
+          B.putLazyByteString bs
+          B.putLazyByteString (BSL.replicate (padLen strSize) 0)
+
+        strs :: [BSL.ByteString] -> B.Put
+        strs = mapM_ str
+
+
+------------------------------------------------------------------------------
+-- | Deserialize a Nar from lazy ByteString
 getNar :: B.Get Nar
 getNar = fmap Nar $ header >> parens getFile
-    where header   = trace "header " $ assertStr "nix-archive-1"
+    where
 
-          padLen n = let r = n `mod` 8
-                         p = (8 - n) `mod` 8
-                     in trace ("padLen: " ++ show p) p
+      header   = assertStr "nix-archive-1"
 
-          str = do
-              n <- fmap fromIntegral B.getInt64le
-              s <- B.getLazyByteString n
-              p <- B.getByteString (padLen $ fromIntegral n)
-              traceShow (n,s) $ return s
 
-          assertStr s = trace ("Assert " ++ show s) $ do
-              s' <- str
-              if s == s'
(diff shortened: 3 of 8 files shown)


hnix-store-core/README.md
```diff
@@ -6,4 +6,12 @@ Core effects for interacting with the Nix store.
 See `StoreEffects` in [System.Nix.Store] for the available operations
 on the store.
 
+[System.Nix.Nar]: ./src/System/Nix/Nar.hs
```
shlevy: If these aren't referenced anywhere in the README no need to add them.

hnix-store-core/hnix-store-core.cabal
```diff
@@ -17,11 +17,35 @@ extra-source-files:  ChangeLog.md, README.md
...
+                     , System.Nix.Path
+                     , System.Nix.Store
   build-depends:       base >=4.10 && <4.11,
                        -- Drop foundation when we can drop cryptonite <0.25
-                       cryptonite, memory, foundation, basement,
+                       binary,
+                       bytestring, containers, cryptonite, memory, foundation, basement,
                        text, regex-base, regex-tdfa-text,
-                       hashable, unordered-containers, bytestring
+                       hashable, unordered-containers, bytestring
```
shlevy: Trailing space?

hnix-store-core/hnix-store-core.cabal
```diff
@@ -17,11 +17,35 @@ extra-source-files:  ChangeLog.md, README.md
...
+     , base
+     , base64-bytestring
+     , binary
+     , bytestring
+     , containers
+     , tasty
+     , tasty-hunit
+     , tasty-hspec
+     , tasty-discover
+   default-language: Haskell2010
```
shlevy: Trailing newline

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,169 @@
...
+import qualified Data.Text.Encoding         as E
+import           GHC.Int                    (Int64)
+
+import           System.Nix.Path
+
+
+data NarEffects (m :: * -> *) = NarEffets {
+    readFile         :: FilePath -> m BSL.ByteString
+  , listDir          :: FilePath -> m [FileSystemObject]
+  , narFromFileBytes :: BSL.ByteString -> m Nar
```
shlevy: Why do we need this as an effect? It's just a pure function right?

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,169 @@
...
+  , narFromDirectory :: FilePath -> m Nar
```
shlevy: Can't we define this with readFile and listDir?

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+
+-- Directly taken from Eelco thesis
+-- https://nixos.org/%7Eeelco/pubs/phd-thesis.pdf
+
+-- TODO: Should we use rootedPath, validPath rather than FilePath?
```
shlevy: Definitely not, you want to create nar files of arbitrary paths, not just valid store paths.

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,169 @@
...
+data Nar = Nar { narFile :: FileSystemObject }
+    deriving (Eq, Ord, Show)
+
+data FileSystemObject =
+    Regular IsExecutable BSL.ByteString
+  | Directory (Set.Set (PathName, FileSystemObject))
```
shlevy: Hmm sets are strict in their values right? So we'd have to recurse all the way through a potentially deeply nested directory to get a top level FSO referring to a dir?

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,169 @@
...
+  | SymLink BSL.ByteString
+  deriving (Eq, Show)
+
+-- TODO - is this right? How does thesis define ordering of FSOs?
```
shlevy: Are we using comparison anywhere in the definition of serialization? Or is this just for `Set`?

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,169 @@
...
+               then str "executable" <> str ""
+               else return ())
+            >> str "contents" <> str contents
+
+        putFile (SymLink target) =
+               str "type" >> str "symlink" >> str "target" >> str target
+
+        putFile (Directory entries) =
+               str "type" <> str "directory"
+            <> foldMap putEntry entries
```
shlevy: So it looks like we're using the `ord` here? I think the right ordering in this case is lexicographical sort on the entry name.

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,169 @@
...
+            parens (str "name" <>
+                    str (BSL.fromStrict $ E.encodeUtf8 name) <>
+                    str "node" <>
+                    parens (putFile fso))
+
+        parens m = str "(" >> m >> str ")"
+
+        str :: BSL.ByteString -> B.Put
+        str t = let len = BSL.length t
+            in int len <> pad t
```
shlevy: Hmm... I'm wondering if lazy IO is insufficient here, if we need to send the len before we send the body. Seems like we'll run into the same issues the C++ code has, where it has to dump the entire thing into memory before it can do anything useful. We might instead need to represent a file as a size + a bytestring, where the size is determined from fstat or whatever, and if the bytestring turns out not to be long enough or too long error out.

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,169 @@
...
+
+        pad :: BSL.ByteString -> B.Put
+        pad bs = do
+          B.putLazyByteString bs
+          B.putLazyByteString (BSL.replicate (padLen (BSL.length bs)) '\NUL')
+
+
+------------------------------------------------------------------------------
+-- | Deserialize a Nar from lazy ByteString
+getNar :: B.Get Nar
```
shlevy: Does this let us operate lazily over the Nar? Or do we have to consume the whole bytestring to be able to do anything with the Nar?

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,169 @@
...
+    compare (Directory _)  _              = GT
+    compare (SymLink l1) (SymLink l2)     = compare l1 l2
+
+data IsExecutable = NonExecutable | Executable
+    deriving (Eq, Show)
+
+
+------------------------------------------------------------------------------
+-- | Serialize Nar to lazy ByteString
+putNar :: Nar -> B.Put
```
shlevy: Hmm... For this to be efficient our nar would have to be constructed with unsafeInterleaveIO. Not necessarily a problem, but I'm wondering if that's the API we want to expose. Maybe it's the simplest for now.

hnix-store-core/src/System/Nix/Nar.hs
shlevy: Thinking more on this I think we definitely want a (lazy) map here.

hnix-store-core/src/System/Nix/Path.hs
```diff
@@ -0,0 +1,76 @@
+{-|
```
shlevy: Is this just copied verbatim or any changes along the way?

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,169 @@
...
+import qualified Data.Binary.Put            as B
+import qualified Data.ByteString.Lazy.Char8 as BSL
+import           Data.Maybe                 (fromMaybe)
+import           Data.Monoid                ((<>))
+import qualified Data.Set                   as Set
+import qualified Data.Text                  as T
+import qualified Data.Text.Encoding         as E
+import           GHC.Int                    (Int64)
+
+import           System.Nix.Path
```
shlevy: Do you actually use this?

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,169 @@
...
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections       #-}
+{-# LANGUAGE TypeApplications    #-}
+
+{-|
+Description : Allowed effects for interacting with Nar files.
+Maintainer  : Shea Levy <shea@shealevy.com>
+|-}
+module System.Nix.Nar where
```
shlevy: Should we have an explicit export list? Or do we want to export `padLen`?

hnix-store-core/src/System/Nix/Store.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE PackageImports #-}
```
shlevy: Why?

hnix-store-core/src/System/Nix/Store.hs
```diff
@@ -110,6 +66,7 @@ data StoreEffects rootedPath validPath m =
       derivationOutputNames :: !(validPath -> m (HashSet Text))
     , -- | Get a full 'Path' corresponding to a given 'Digest'.
       pathFromHashPart :: !(Digest PathHashAlgo -> m Path)
+    , narEffects :: NarEffects m
```
shlevy: We can move this around later, but I'm not sure it necessarily needs to be here. The `addNarToStore` effect can just take a `Nar`. We'll see.

hnix-store-core/tests/Nar.hs
```diff
@@ -0,0 +1,150 @@
+{-# LANGUAGE OverloadedStrings #-}
```
shlevy: Hmm I feel like nar serialization/deserialization is a perfect case for property testing (quickcheck or whatever). Would you mind giving that a spin?

shlevy: Maybe in addition to tests derived from seeing what the nix cli does.

shlevy: Actually maybe we can just for now assume that the CLI is available and run it during the tests, instead of pre-generating fixtures.

domenkozar:
Couldn't this be a completely standalone package? It doesn't depend on any Nix related haskell libs :)

shlevy:
hnix-store-core is pretty independent already...

hnix-store-core/src/System/Nix/Nar.hs
imalsogreg: Ahh, yes. A hashmap from name to FSO makes more sense than `Set (PathName, FSO)`, since names must be unique. I'll change to that.

hnix-store-core/src/System/Nix/Nar.hs
imalsogreg: Great point! The Haskell [tar](https://hackage.haskell.org/package/tar) package claims to do laziness correctly - I'll have to read through the code to see if I can find any tips. `BS.length` will definitely force the whole bytestring, so I like the fstat (or whatever :) ) idea.

hnix-store-core/src/System/Nix/Nar.hs
imalsogreg: IIUC, when i get the implementation right, we can go through `Nar` lazily streaming one from network <-> disk, or pulling out a individual entries in constant space. But the implementation right now is definitely not going to achieve that.

hnix-store-core/src/System/Nix/Nar.hs
imalsogreg: Since we are producing a lazy bytestring, can't `putNar` run in constant space? Only if we forced the `Nar` would I expect this approach to blow up. Well, my intuition here is completely un-tested :) so...

hnix-store-core/src/System/Nix/Path.hs
imalsogreg: Definitions copied verbatim, but I think I changed the exports list.

hnix-store-core/src/System/Nix/Nar.hs
imalsogreg: Yep, in the `Directory` constructor for `FileSystemObject` I use `PathName`.

hnix-store-core/src/System/Nix/Nar.hs
imalsogreg: I'll make the export list. `padLen` is shared between `put` and `get` but otherwise is an internal implementation detail.

hnix-store-core/src/System/Nix/Store.hs
imalsogreg: Leftover, will remove.

hnix-store-core/tests/Nar.hs
imalsogreg: I'll add an `Arbitrary` sized instance and some tests that shell out to `nix-store`. Will keep the hand-copied fixtures for when "we win" and there is no more reference `nix-store` to use for tests :P

imalsogreg: Actually, I'm stumped about how you could _ever_ encode a string's length, followed by the string itself, without forcing the full string....

But looking at the source for `ByteString`, things look promising! Strict bytestrings give `length` in O(1), and it doesn't look like it forces the underlying bytestring. https://hackage.haskell.org/package/bytestring-0.10.8.2/docs/src/Data.ByteString.html#length ... and in the lazy case, `BSL.length bsl` is defined with a foldl' over the chunks ( https://hackage.haskell.org/package/bytestring-0.10.8.2/docs/src/Data.ByteString.Lazy.html#length ), which is strict in the `Int64` being accumulated, but shouldn't force the whole of `bsl`.... I'm not really sure I get the laziness going on here correctly, need to do an experiment.

hnix-store-core/src/System/Nix/Nar.hs
shlevy: To have a `Nar` in the first place requires us to have already recursed all the directories and read all the files, unless we're interleaving IO, right? How else would laziness save us here?

hnix-store-core/tests/Nar.hs
shlevy: @imalsogreg You can do it by trusting the filesystem to correctly report the size of the file being read, and hoping the file doesn't change in the mean time (if it does, then you need to just error out in a streaming context)

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,216 @@
...
+
+-- Directly taken from Eelco thesis
+-- https://nixos.org/%7Eeelco/pubs/phd-thesis.pdf
+
+data Nar = Nar { narFile :: FileSystemObject }
+    deriving (Eq, Show)
+
+data FileSystemObject =
+    Regular IsExecutable BSL.ByteString
+  | Directory (Map.Map PathName FileSystemObject)
```
shlevy: `PathName` is incorrect here. `PathName` is limited to just what Nix allows in the *top-level* name of an output, but files within that can be named anything. The real rule here is an arbitrary byte-string with no nulls or `/`s.

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,216 @@
...
+              else fail "No"
+
+
+-- | Distance to the next multiple of 8
+padLen :: Int64 -> Int64
+padLen n = (8 - n) `mod` 8
+
+
+-- | Unpack a FileSystemObject into a non-nix-store directory (e.g. for testing)
+localUnpackNar :: FilePath -> Nar -> IO ()
```
shlevy: This is awesome, but I'm not sure it belongs here, as we're trying to be effect-agnostic. Can this be implemented in terms of a coherent set of filesystem effects? Or if not maybe we just move this to a module giving us the plain POSIX filesystem implementation of `NarEffects`?

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,216 @@
...
+localUnpackNar basePath (Nar fso) = do
+
+  -- Create regular files and directories,
+  -- But save link creation until the end, by writing link commands out
+  -- to a queue, since otherwise we may try to create a link before
+  -- its target exists
```
shlevy: Wait, why does this matter? symlinks don't need to be valid, you can absolutely have a nar with `foo -> bar` but nothing at `bar`.

shlevy:
@imalsogreg ping me when this is ready for fresh review or if you have any questions!

imalsogreg:
@shlevy Ping!

New Arbitrary instance and shell-out-to-`nix store --dump` provide a couple more test cases:

```
[nix-shell:~/code/hnix-store/hnix-store-core]$ dist/build/format-tests/format-tests
tests/Driver.hs
  narEncoding
    parser-roundtrip
      roundtrips regular:    OK
      roundtrips regular 2:  OK
      roundtrips executable: OK
      roundtrips symlink:    OK
      roundtrips directory:  OK
    matches-nix-store fixture
      matches regular:       OK
      matches regular':      OK
      matches executable:    OK
      matches symlink:       OK
      matches directory:     OK
  nixStoreRegular:           OK (0.04s)
  nixStoreDirectory:         OK (0.03s)
  nixStoreDirectory':        OK (0.03s)
  narEncodingArbitrary:      OK (0.04s)
    +++ OK, passed 100 tests.

All 14 tests passed (0.16s)
```

I would very much like to test/optimize the laziness, but I think that may be a good thing for a second PR. But, up to your preference!

imalsogreg:
@shlevy actually there are a couple things in your prior comments left for me to do. I'll update the TODO list. Feel free to edit it if you would like go grow or shrink the scope for _this_ PR :)

shlevy:
@imalsogreg For scope cutting, I think we can move actually computing a store hash and IO instances out of this initial PR... For me the main thing at this stage is just avoiding the memory explosion seen in C++ Nix :smile: 

What is the binary instance used for again?

shlevy:
Note that "computing the store hash" is in principle a `StoreEffect`, as when we talk to the daemon there's no real need to compute it in advance, we can just stream the nar to the daemon and it will give us a path back. We'll need a haskell implementation though for readonly and mock mode.

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,231 @@
...
+import qualified Data.Text.Encoding         as E
+import           GHC.Int                    (Int64)
+import           System.Directory
+import           System.FilePath
+import           System.Posix.Files         (createSymbolicLink)
+
+import           System.Nix.Path
+
+
+data NarEffects (m :: * -> *) = NarEffects {
```
shlevy: Maybe add a TODO for separating out read effects from write effects. Certainly not necessary to start though.

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,231 @@
...
+    narReadFile   :: FilePath -> m BSL.ByteString
+  , narWriteFile  :: FilePath -> BSL.ByteString -> m ()
+  , narListDir    :: FilePath -> m [FilePath]
```
shlevy: Technically these can't be just any file paths, as they can't contain directory separators, and really it's a set not a list. Also, on most filesystems, readdir will return whether the file is a directory or not in the same call, which may be worth exposing at some point.

hnix-store-core/src/System/Nix/Nar.hs
imalsogreg: Done

hnix-store-core/src/System/Nix/Nar.hs
imalsogreg: The nar may contain invalid links, but we have to sequence creation of those links on the disk, since the trying to _create_ a link will fail if the target does not exist.

Or so I thought!! TIL that `ln -s path1 path2` does not check that `path1` exists! So I'll write `localUnpackNar` without the `WriterT` part.

imalsogreg:
@shlevy So there is enough code in place now to decode/encode NARs in at least the naive way. I'd like to get some more help figuring out how to test the main use cases. So what do you think about the sequence: (1) merge this PR as is, (2) open issues about perf/laziness testing (3) open issues about changing the API (either to make it fit the domain better, or changing it for performance reasons)?

`binary` instance here is just to connect the names `getNar` and `putNar` to the instance methods `get` and `put` of the class `Binary`. So, finishing that definition was only a matter of: `instance Binary Nar where; get = getNar; put = putNar`.

shlevy:
Will give this another round of review this weekend.

imalsogreg:
Ok, even after replacing `BSL.length bs <> str bs` style (probably-not-lazy-enough) encoding decoding with a version that should not force the strings, an call to `packNar >>= writeBSL` will load the whole file into memory. I'll try to rewrite with `streaming`.

imalsogreg:
Now, tests can be run with an environment variable that gives large sizes to temporary files. Combined with RTS  options that limit heap size, we can prove that `Nar`s can serve as intermediate data structures when we streaming nar encoded bytestrings to unpacked filesystems and vice versa

`~/code/hnix-store/hnix-store-core]$ HNIX_BIG_FILE_SIZE=1000000000 dist/build/format-tests/format-tests +RTS -M100M `

This produces some 1G single files and directories with 1G files in the test suite, while locking the heap size to 100M. With this combination of parameters, and the recent change in encoding/decoding strategy (being more careful not to force the intermediate lazy bytestrings) the tests pass. The tests can be made to fail by shrinking heap size below 100M.

domenkozar:
Small tip if it's helpful how to check max memory within a single test: https://github.com/plow-technologies/servant-streaming/blob/master/servant-streaming-server/test/Servant/Streaming/ServerSpec.hs#L83-L84

hnix-store-core/tests/NarFormat.hs
```diff
@@ -0,0 +1,354 @@
...
+
+-- | Add a link to a FileSystemObject. This is useful
+--   when creating Arbitrary FileSystemObjects. It
+--   isn't implemented yet
+mkLink ::
+     FilePath -- ^ Target
+  -> FilePath -- ^ Link
+  -> FileSystemObject -- ^ FileSystemObject to add link to
+  -> FileSystemObject
+mkLink = undefined -- TODO
```
shlevy: This isn't referenced anywhere, should we remove it?

hnix-store-core/LICENSE
```diff
@@ -1 +0,0 @@
-../LICENSE
```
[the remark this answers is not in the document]
imalsogreg: When LICENSE was a symlink, I had trouble bringing `hnix-store-core` in as an override within `hnix`. Short-term stopgap. Want me to revert to the symlink?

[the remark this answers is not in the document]
shlevy: Please, unless this is too difficult to work around

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,261 @@
...
+import           Data.Traversable           (forM)
+import           GHC.Int                    (Int64)
+import           System.Directory
+import           System.FilePath
+import           System.Posix.Files         (createSymbolicLink, fileSize, getFileStatus,
+                                             isDirectory, readSymbolicLink)
+
+import           System.Nix.Path
+
+data NarEffects (m :: * -> *) = NarEffects {
```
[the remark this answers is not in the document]
imalsogreg: Interface design is definitely going to be up to your taste. I'd rather get something committed and defer experimentation to later PRs. Sure, we can discuss further in a call.

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,261 @@
...
+            in int len <> pad len t
+
+        putContents :: Int64 -> BSL.ByteString -> B.Put
+        putContents fSize bs = str "contents" <> int fSize <> (pad fSize bs)
+        -- putContents fSize bs = str "contents" <> int (BSL.length bs) <> (pad fSize bs)
+
+        int :: Integral a => a -> B.Put
+        int n = B.putInt64le $ fromIntegral n
+
+        pad :: Int64 -> BSL.ByteString -> B.Put
```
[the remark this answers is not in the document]
imalsogreg: Yah. May be tough to do this without causing a space leak. New ticket for trying that out?

hnix-store-core/src/System/Nix/Nar.hs
```diff
@@ -0,0 +1,261 @@
...
+          assertStr "type"
+          assertStr "symlink"
+          assertStr "target"
+          fmap (SymLink . E.decodeUtf8 . BSL.toStrict) str
+
+      getEntry = do
+          assertStr "entry"
+          parens $ do
+              assertStr "name"
+              mname <- E.decodeUtf8 . BSL.toStrict <$> str
```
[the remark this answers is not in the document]
imalsogreg: Used to be a `Maybe`. Fixed name no reflect pureness.

hnix-store-core/tests/NarFormat.hs
imalsogreg: Rather keep it as a TODO. It's a reminder to come up with a sensible way to add symlinks to the arbitrary instance.

PR #11, opened 2018-05-14
README: Elaborate on project goals and layout.

### LINKED ISSUE #9: Elaborate on project goals and package structure in README.md
Could you give some more info - both in terms of API boundary and specific examples - of what `hnix-store` is meant to do? I've heard you say that we want to communicate with the nix daemon, but I'm not sure if that is the high level goal of the project, a design constraint for the whole project, or a design constraint for a subset of the packages within the project.

The ideal description would target people with only a vague understanding of the nix daemon and its relationship to `nix-store`, `nix-instantiate` and `nix-env` - mainly as an aid to contributors who know more about Haskell than Nix, like yours truly. :)

shlevy:
Fixes #9.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,64 @@ A Haskell interface to the [Nix] store.
 
 [Nix]: https://nixos.org/nix
 
+Rationale
+-----------
+
+`Nix` can conceptually be broken up into two layers, both (perhaps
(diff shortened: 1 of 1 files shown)


shlevy:
@imalsogreg Please take a look and review

Issue #12, opened 2018-05-16
Improve NarEffects and API


imalsogreg:
We should define an API that's suitable for use by `hnix`, and also appropriate for use of `hnix-store` as a CLI tool.

This includes cleaning up `NarEffects` record (maybe turning it into a typeclass to match style with `hnix`), and defining other `NarEffects` implementations suitable for `readonly` and `mock` mode.

We should try to dogfood this soon by calling `hnix-store-core` from `hnix`, and letting that drive more API improvements. For example, we should be able to easily support https://github.com/haskell-nix/hnix/issues/299

PR #13, opened 2018-07-16
WIP: Remote store

sorki:
Prototype of remote store access via worker protocol. Most of the API is implemented, tho I'm struggling a bit with path hash computation/extraction. Few methods are still left undefined as they require some special care.

I'm not quite sure how to fit this onto API in `System.Nix.Store` as well as it differentiates between `rootedPath` and `validPath` - not sure how are these related to `Path` and `ValidPathInfo` (also `ValidPathInfo` needs separate file due to clashing record field names).

Related to #12 and haskell-nix/hnix#173

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,0 +1,1 @@
+packages: ./hnix-store-core/*.cabal ./hnix-store-remote/*.cabal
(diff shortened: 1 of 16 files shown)


sorki:
I will also add `ErrorT` handling as per notes but wanted to get this out as it's a lot of code already.

shlevy:
Thanks! @imalsogreg are you able to give this a first-pass review?

Issue #4, opened 2018-04-28
Investigate using cryptohash-sha256 instead of cryptonite


mightybyte:

imalsogreg:
I believe we lose 2 things by moving to `cryptohash-sha256`, and that the benefit we get from them doesn't justify the cost of using `cryptonite`:

 - Type-level tracking of the hash algorithm. There are a limited number ( < 10 ) of ways that hashes are computed, and we can just have a newtype wrapper for each. Or steal the `Digest` type family from `cryptonite`.
 - Base32 encoding of a sha256 digest. I'm not sure how maintained the [base32-bytestring](https://hackage.haskell.org/package/base32-bytestring) library is, plus nix uses its own unique character mapping (it's incompatible with `cryptonite`'s, but we could probably Caesar-cypher between the two)... so maybe we shoud just implement that ourselves, [as nix does](https://github.com/NixOS/nix/blob/master/src/libutil/hash.cc#L61).  To get a gist of what this would look like, see my [cryptohash-256 branch of hnix-store](https://github.com/imalsogreg/hnix-store/compare/master...imalsogreg:cryptohash-256), where I've gotten started trying to do that. No testing or careful thought about corner cases has been done yet, WIP.

imalsogreg:
We gain a smaller dependency footprint (`basement`, `foundation`, `memory` and `cryptonite` go away). I'm not sure if truncation is a formally defined thing, or simply means "take all but the last N bytes" - do you expect it to be hard?

shlevy:
Nah, not hard, just was pleasantly surprised at how straightforward and generic the cryptonite version is. Anyway, happy to accept a PR switching

domenkozar:
Relevant: https://github.com/vincenthz/hs-memory/issues/67

imalsogreg:
Closed by #27

PR #14, opened 2018-08-05
Cryptohash sha256

imalsogreg:
- Add nix-style base32 encoder from @ryantrinkle
 - Drop `Crypto.Hash.Truncated` module
 - Minor haddock comment fixup
 - Drop dependency on `cryptonite` package

Addresses #4


shlevy (changes requested):
Recording here for posterity: Let's make `StorePathHash` a properly encapsulated newtype, constructed with the normal digest operations

hnix-store-core/src/System/Nix/Hash.hs
```diff
@@ -0,0 +1,105 @@
+{-|
+Description : Nix-style hashes (truncated sha256)
+Maintainer  : Shea Levy <shea@shealevy.com>
+-}
+
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+module System.Nix.Hash (
+    StorePathHash (..)
```
shlevy: Do we want to export the constructor?

hnix-store-core/src/System/Nix/Hash.hs
```diff
@@ -0,0 +1,105 @@
...
+
+hashlazy :: BSL.ByteString -> StorePathHash
+hashlazy bs = StorePathHash . truncate52 $ SHA.hashlazy bs
+
+
+-- | Import and validate a store path hash
+fromBase64Digest :: BS.ByteString -> Maybe StorePathHash
+fromBase64Digest = validateRawDigest . StorePathHash
+  where validateRawDigest = Just
+  -- TODO: What should we check for? Only valid base64 chars?
```
shlevy: base64? Do you mean base32? Anyway, we should check that it's of the right length and only has the right charcters.

Wait, why are we putting either of these things into a StorePathHash? That's just a 160 bit bytestring, right? No characters of any kind.

hnix-store-core/src/System/Nix/Hash.hs
```diff
@@ -0,0 +1,105 @@
...
+  -- Truncate 52 bits by dropping 6 bytes worth of Word8's,
+  -- then masking 4 bits off of the 7th Word8
+  case BS.uncons (BS.drop (52 `div` 8) digest) of
+    Nothing     -> BS.empty -- We received an hash with unexpectedly short length
+    Just (x,xs) -> BS.cons (mask4bits .&. x) xs
+    where mask4bits = 2^5 - 1 :: Word8
+
+
+-- | Convert a ByteString to base 32 in the way that Nix does
+toNixBase32 :: BSL.ByteString -> BSL.ByteString
```
shlevy: Is the input bytestring expected to be 160 bits, or ascii represenation of hex, or something else?

Issue #24, opened 2018-09-12
base32 hashes do not match those produced by nix-store


imalsogreg:

shlevy:
Can you show some examples? Note that there are two ways to hash a file, `flat` hashes are just traditional hashes like what you'd get from `sha256sum` (though converted to base32) or `recursive` hashes which do a nar dump even for a single file.

imalsogreg:
Closed by #28

PR #25, opened 2018-09-12
Refactor System.Nix.Hash to drop cryptonite dependency

imalsogreg:
This PR supersedes PR #14 by moving hashing operations into a module and dropping `cryptonite`, `basement` and `foundation` dependencies, to facilitate adding `hnix-store` as a dependency of `hnix`.

Test of string hashing still fail, and we may want `StorePathHash` to encapsulate a raw bytestring, rather than the truncated, base32 encoded version - but maybe we should make those changes together in a new PR?

String hashing test failure is tracked in https://github.com/haskell-nix/hnix-store/issues/24


hnix-store-core/src/System/Nix/Hash.hs
```diff
@@ -0,0 +1,105 @@
...
+import           Prelude                      hiding (init)
+
+-- | A string, file, or NAR hash in the format
+--   used in prefixing files in the nix store
+newtype StorePathHash =
+  StorePathHash { getTruncatedHash :: BS.ByteString }
+  deriving (Eq, Hashable.Hashable, Ord, Show)
+
+
+init :: SHA.Ctx
```
shlevy: Can we encapsulate the context type here?

hnix-store-core/src/System/Nix/Hash.hs
```diff
@@ -0,0 +1,105 @@
...
+hash :: BS.ByteString -> StorePathHash
+hash bs = StorePathHash . BSL.toStrict . toNixBase32 . BSL.fromStrict . truncate' $ SHA.hash bs
+
+
+hashlazy :: BSL.ByteString -> StorePathHash
+hashlazy bs = StorePathHash . BSL.toStrict . toNixBase32 . BSL.fromStrict . truncate' $ SHA.hashlazy bs
+
+
+-- | Import and validate a store path hash
+fromBase32 :: BS.ByteString -> Maybe StorePathHash
```
shlevy: base 32/base 16 questions should just be about presentation, not about internal represenation

hnix-store-core/tests/Hash.hs
```diff
@@ -0,0 +1,22 @@
...
+import           Data.Semigroup
+import           Test.Tasty.Hspec
+import           Test.Tasty.HUnit
+import           System.Nix.Hash
+
+spec_hashBase32truncateParity :: Spec
+spec_hashBase32truncateParity = describe "hashBase32" $
+  for_ testCases $ \(testCase, expectation) ->
+    it ("computes correct base32 hash for string " <> BSC.unpack testCase) $
+    getTruncatedHash (hash testCase) `shouldBe` expectation
```
shlevy: No it shouldn't :wink:

hnix-store-remote/src/System/Nix/Store/Remote.hs
```diff
@@ -148,19 +148,19 @@ queryDerivationOutputNames p = do
   sockGetPaths
 
 -- XXX: this is broken as I don't know how to get hashes from paths (fix mkPath)
-queryPathFromHashPart :: Digest PathHashAlgo -> MonadStore (Maybe Path)
+queryPathFromHashPart :: StorePathHash -> MonadStore (Maybe Path)
 queryPathFromHashPart d = do
   runOpArgs QueryPathFromHashPart $
-    putByteStringLen $ LBS.fromStrict $ convert d
+    putByteStringLen $ LBS.fromStrict $ getTruncatedHash d
```
shlevy: *here* we do need to convert to base32, since we're talking to the daemon.

hnix-store-remote/src/System/Nix/Store/Remote.hs
```diff
@@ -148,19 +148,19 @@ queryDerivationOutputNames p = do
...
   sockGetPath
 
 type Source = () -- abstract binary source
 addToStoreNar :: ValidPathInfo -> Source -> RepairFlag -> CheckSigsFlag -> MonadStore ()
 addToStoreNar = undefined  -- XXX
 
 type PathFilter = Path -> Bool
-addToStore :: LBS.ByteString -> Path -> Bool -> PathHashAlgo -> PathFilter -> RepairFlag -> MonadStore Path
-addToStore name pth recursive hashAlgo pfilter repair = undefined -- XXX
+addToStore :: LBS.ByteString -> Path -> Bool -> PathFilter -> RepairFlag -> MonadStore Path
```
shlevy: Why is this argument dropped? We should be able to pass different hash algorithms to this operation, we can add things to the store based on e.g. md5 or sha512 hash

shlevy (changes requested):
Seem to be a few confusions around base32 handling

Issue #18, opened 2018-09-11
Align hnix-store dependencies with hnix


shlevy:
We should trim down our dependencies to make it straightforward for hnix to depend on us

imalsogreg:
`hnix-store-core` and `hnix-store-remote` have very light dependencies, except for `cryptonite` and `cryptonite`'s dependencies. 

I've removed those, in PR https://github.com/haskell-nix/hnix-store/pull/25

Anton-Latukha:
Noted that HNix Nix closure compiles `basement`, `foundation`, `memory`.
And indeed.
Looked into it: https://hydra.nixos.org/build/132735611#tabs-build-deps

The HNix includes the whole shebang with transitive dependencies. Basically through `http-client{,-tls}` which use `memory`.

HNix currently temporarily uses `hashing` for convenience. Discovered that `hashing` in reality uses `cryptonite`. IDK through what - it is not indicated in the deps, but the Nix closure and https://hydra.nixos.org/build/132735611#tabs-build-deps clearly indicate that package depends on `cryptonite`.

Because `cryptohash-*` become unmaintained, I started the Hash subsystem refactoring to switch it to `cryptonite`, I think currently there is no escaping the switch to the current standard library in Haskell.

Meta thread about it is: https://github.com/haskell-nix/hnix-store/issues/92

Anton-Latukha:
After the #107 and #117 the deps are minimal.

And aligned with HNix. Except for hash, which in both places needs to be switched asap, and there is no alternatives to `cryptonite`.

PR #27, opened 2018-09-30
Remove cryptonite, foundation, basement and memory from all packages

imalsogreg:
Addresses #18 by dropping `cryptonite` and related libraries from `hnix-store-core` and `hnix-store-remote`.

This PR supersedes #25 and #14, both of which tried to add correct path hashing. But we need to do more planning and design before we are ready to implement that. This PR does the minimal work necessary to remove the `cryptonite` dependency while continuing to allow `hnix-store-core` and `hnix-store-remote` to compile, ignoring path hashes for the moment.

The change to `System.Nix.Store.Remote.Util.mkPath` gives us a loss of functionality (a runtime error), until we implement path hashing ourselves. The previous behavior likely generated a hash that does not match that produced by `nix-store`.


shlevy:
Sorry @imalsogreg busy week, will look at this soon!

imalsogreg:
@shlevy no hurry, not blocked :)

hnix-store-core/src/System/Nix/Hash.hs
```diff
@@ -0,0 +1,30 @@
...
+import Data.Hashable (Hashable (..))
+import Data.Proxy (Proxy(..))
+import Data.Word (Word8)
+import GHC.TypeLits (Nat, KnownNat, natVal, type (<=))
+import Foreign.Ptr (castPtr, Ptr)
+import Foreign.Marshal.Utils (copyBytes)
+
+data HashAlgorithm = TruncatedSHA256 | MD5
+
+newtype Digest (algo :: HashAlgorithm) = Digest { getDigestBytes :: BS.ByteString }
```
shlevy: I'm confused... How are we meant to construct a Digest? Where is truncation happening?

imalsogreg: This is just a placeholder until we have an actual API design for Digests. There is no code here for producing a digest because the scope of this PR is limited to simplifying the dependency list and reintroducing a small number of stubs to continue allowing things to compile.

The actual Digest API, which supersedes this one, is in https://github.com/haskell-nix/hnix-store/pull/28, which I think should be reviewed separately since the scope is much larger.

Moving forward with #27 and doing fixup in #28 seems harmless, because all the digest stuff in play prior to #28 is just stubs anyway, but I'm open to alternatives.

PR #28, opened 2018-10-22
Add basic hashing API

imalsogreg:
- [x] Internal module for data definitions and implementation
  - Everything is exported from `.Internal.Hash` ( "internal-by-convention" for unsafe constructors and functions )
 - [x] `System.Nix.Hash` reexports only the public-facing parts of `.Internal.Hash`
 - [x] `initialize`/`update`/`finalize` available for MD5, SHA-1, SHA-256
 - [x] `hash`/`hashLazy` functions provided for hashing bytestrings in one go (as in cryptonite)
 - [x] Tests (We have 2 tiny unit tests, one for `SHA1 -> base32`, one for `SHA256 -> truncated base32`)

### DIFF
--- a/hnix-store-core/src/System/Nix/Internal/Hash.hs
+++ b/hnix-store-core/src/System/Nix/Internal/Hash.hs
@@ -1,0 +1,166 @@
+{-|
+Description : Cryptographic hashes for hnix-store.
+Maintainer  : Greg Hale <imalsogreg@gmail.com>
+-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE PolyKinds           #-}
+{-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE TypeFamilies        #-}
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
+
+module System.Nix.Internal.Hash where
+
+import qualified Crypto.Hash.MD5       as MD5
+import qualified Crypto.Hash.SHA1      as SHA1
+import qualified Crypto.Hash.SHA256    as SHA256
+import qualified Data.ByteString       as BS
+import qualified Data.ByteString.Char8 as BSC
+import           Data.Bits             (xor)
+import qualified Data.ByteString       as BS
+import qualified Data.ByteString.Lazy  as BSL
+import qualified Data.Hashable         as DataHashable
+import           Data.List             (foldl')
+import           Data.Proxy            (Proxy(Proxy))
+import qualified Data.Text             as T
+import qualified Data.Text.Encoding    as T
+import qualified Data.Vector           as V
+import           Data.Word             (Word8)
+import           GHC.TypeLits
+
+-- | A tag for different hashing algorithms
+--   Also used as a type-level tag for hash digests
+--   (e.g. @Digest SHA256@ is the type for a sha256 hash)
+data HashAlgorithm
+  = MD5
+  | SHA1
+  | SHA256
+  | Truncated Nat HashAlgorithm
+
+
+-- | Types with kind @HashAlgorithm@ may be a @HasDigest@ instance
+--   if they are able to hash bytestrings via the init/update/finalize
+--   API of cryptonite
+--
+--   Each instance defined here simply defers to one of the underlying
+--   monomorphic hashing libraries, such as `cryptohash-sha256`.
+class HasDigest (a :: HashAlgorithm) where
+
+  type AlgoCtx a :: *
+
+  initialize        :: AlgoCtx a
+  update            :: AlgoCtx a -> BS.ByteString -> AlgoCtx a
+  finalize          :: AlgoCtx a -> Digest a
+
+
+-- | The cryptographic hash of of a strict bytestring, where hash
+--   algorithm is chosen by the type of the digest
+--   For example:
+--   > let d = hash "Hello, sha-256!" :: Digest SHA256
+--   or
+--   > :set -XTypeApplications
+--   > let d = hash @SHA256 "Hello, sha-256!"
+hash :: forall a.HasDigest a => BS.ByteString -> Digest a
+hash bs =
+  finalize $ update @a (initialize @a) bs
+
+-- | The cryptographic hash of a lazy bytestring. Use is the same
+--   as for @hash@. This runs in constant space, but forces the
+--   entire bytestring
+hashLazy :: forall a.HasDigest a => BSL.ByteString -> Digest a
+hashLazy bsl =
+  finalize $ foldl' (update @a) (initialize @a) (BSL.toChunks bsl)
+
+
+
+
+-- | Convert any Digest to a base32-encoded string.
+--   This is not used in producing store path hashes
+printAsBase32 :: Digest a -> T.Text
+printAsBase32 (Digest bs) = printHashBytes32 bs
+
+
+instance HasDigest MD5 where
+  type AlgoCtx 'MD5 = MD5.Ctx
+  initialize = MD5.init
+  update = MD5.update
+  finalize = Digest . MD5.finalize
+
+instance HasDigest 'SHA1 where
+  type AlgoCtx SHA1 = SHA1.Ctx
+  initialize = SHA1.init
+  update = SHA1.update
+  finalize = Digest . SHA1.finalize
+
+instance HasDigest 'SHA256 where
+  type AlgoCtx SHA256 = SHA256.Ctx
+  initialize = SHA256.init
+  update = SHA256.update
+  finalize = Digest . SHA256.finalize
+
+instance (HasDigest a, KnownNat n) => HasDigest (Truncated n a) where
+  type AlgoCtx (Truncated n a) = AlgoCtx a
+  initialize = initialize @a
+  update = update @a
+  finalize = truncateDigest @n . finalize @a
+
+-- | A raw hash digest, with a type-level tag
+newtype Digest (a :: HashAlgorithm) = Digest
+  { digestBytes :: BS.ByteString
+    -- ^ The bytestring in a Digest is an opaque string of bytes,
+    --   not some particular text encoding.
+  } deriving (Show, Eq, Ord, DataHashable.Hashable)
+
+
+-- instance DataHashable.Hashable (Digest a) where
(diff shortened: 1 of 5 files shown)


hnix-store-core/src/System/Nix/Internal/Hash.hs
```diff
@@ -0,0 +1,152 @@
...
+--   API of cryptonite
+--
+--   Each instance defined here simply defers to one of the underlying
+--   monomorphic hashing libraries, such as `cryptohash-sha256`.
+class HasDigest (a :: HashAlgorithm) where
+
+  type AlgoCtx a :: *
+  hashLength        :: Proxy a -> Int
+
+  validateHashBytes :: BS.ByteString -> Maybe (Digest a)
```
shlevy: Hmm... Do we need this? We might, but my inclination is to not expose it until we have a use case.

hnix-store-core/src/System/Nix/Internal/Hash.hs
```diff
@@ -0,0 +1,152 @@
...
+  validateHashBytes bs = if BS.length bs == hashLength (Proxy @a)
+                         then Just (Digest bs)
+                         else Nothing
+
+  initialize        :: AlgoCtx a
+  update            :: AlgoCtx a -> BS.ByteString -> AlgoCtx a
+  finalize          :: AlgoCtx a -> Digest a
+
+  toBase32          :: Digest a -> T.Text
```
shlevy: I don't think this is a class method really... Should just be a method on `Digest a`s

hnix-store-core/src/System/Nix/Internal/Hash.hs
```diff
@@ -0,0 +1,152 @@
...
+-- | Types with kind @HashAlgorithm@ may be a @HasDigest@ instance
+--   if they are able to hash bytestrings via the init/update/finalize
+--   API of cryptonite
+--
+--   Each instance defined here simply defers to one of the underlying
+--   monomorphic hashing libraries, such as `cryptohash-sha256`.
+class HasDigest (a :: HashAlgorithm) where
+
+  type AlgoCtx a :: *
+  hashLength        :: Proxy a -> Int
```
shlevy: nit: `proxy a -> Int`.

FWIW I prefer [Tagged](http://hackage.haskell.org/package/tagged-0.8.6/docs/Data-Tagged.html), but I know conventions are against me here

hnix-store-core/src/System/Nix/Internal/Hash.hs
```diff
@@ -0,0 +1,152 @@
...
+  toBase32 (Digest bs) = T.decodeUtf8 $ printHashBytes32 bs
+
+instance HasDigest 'SHA256 where
+  type AlgoCtx SHA256 = SHA256.Ctx
+  hashLength _ = 32
+  initialize = SHA256.init
+  update = SHA256.update
+  finalize = Digest . SHA256.finalize
+  toBase32 (Digest bs) =
+    T.decodeUtf8 $ printHashBytes32 $ truncateDigest 20 bs
```
shlevy: ? We definitely need untruncated sha-256 hashes (and their base 32 representations) as well as truncated ones, and in any case if we're dealing with truncated hashes it's the entire hash that's truncated, not just its human-friendly representation (e.g. when comparing for equality). IMO `HashAlgorithm` should have a `Truncated` constructor taking another `HashAlgorithm`

hnix-store-core/src/System/Nix/Internal/Hash.hs
```diff
@@ -0,0 +1,152 @@
...
+        digitInd = fromIntegral $
+                   sum [fromIntegral (byte j) * (256^j)
+                       | j <- [0 .. BS.length c - 1]]
+                   `div` (32^i)
+                   `mod` 32
+
+
+-- TODO Test this, too
+truncateDigest :: Int -> BS.ByteString -> BS.ByteString
+truncateDigest n c = BS.pack $ map truncOutputByte [0.. n-1]
```
shlevy: Why isn't this just `BS.take`?

imalsogreg: This is a straight (and inefficient!) translation of the thesis. Because `n < |c|` (truncated length is less than the input bytestring length), some positions in the input bytestring get to impact the truncated version twice.

A more straightforward implementation would be: 
```
let layer1  = BS.take n bs
    -- leading bits that aren't truncated
    layer2 = BS.drop n bs <> BS.replicate (BS.length bs - n) 0
    -- ^ trailing bits, right padded
    -- v xor the two together for "truncated" digest
in xor layer1 layer2
``` (if xor could work bitwise over a bytestring)

Comment (shlevy) [reply hnix-store-core/src/System/Nix/Internal/Hash.hs]: Wait, seriously? It's not just a naive truncation? Why???

Comment (shlevy) [reply hnix-store-core/src/System/Nix/Internal/Hash.hs]:
```c++
    Hash h;
    h.hashSize = newSize;
    for (unsigned int i = 0; i < hash.hashSize; ++i)
        h.hash[i % newSize] ^= hash.hash[i];
    return h;
```
OK then

hnix-store-core/src/System/Nix/Internal/Hash.hs
imalsogreg:
```
let h_bytes = hash256(string)
    h_bytes' = truncate(h_bytes)
    forStorePath = toBase32(h_bytes')
```
Imagining I go through these steps to produce a store path hash. There are multiple places in the code where we need the untruncated digest, and multiple places where we need the final base32 encoded truncated hash.

Is there any place in the code other than generating base32 encodings, where we would need access to the truncated version of the hash bytes? If so, then a `Truncated` newtype would be useful for carrying the proof that truncation has happened. OTOH if truncation is merely an implementation detail of producing base32 encoding of a hash, then `Truncated` newtype wouldn't pull its weight. What do you think?

hnix-store-core/src/System/Nix/Internal/Hash.hs
imalsogreg: I'm using the class signal different types going to base32 in different ways (where MD5 and SHA1 simply base32Bytes their hash, while a SHA256 hash needs to do truncation first).

The fact that different types have different `base32` behaviors in my code might be a mistake in my mental model. Like the way I don't have a `Truncated` newtype in the API. If we _did_ have `newtype Truncated = T BS.Bytestring`, then I could see defining a few monomorphic funuctions `base32FromMD5` `base32FromTruncatedSHA256`, `base32FromSHA1`. A polymorphic `base32` function over `Digest a` would be able to produce some strings that aren't used by nix - like `base32 (someHash :: Digest 'SHA256)` <- that would be an "invalid" (not used by nix) `Text`.

imalsogreg: Thesis (Page 95) talks about the `valid : Path -> Hash` mapping, which does use untruncated SHA256 hashes. So yep, have to rethink this :)

imalsogreg:
@shlevy, Addressed all your comments. Especially the one about moving `base32` out of the typeclass, since yes, non-truncated sha256 hashes do sometimes get used in nix-store.

No hurry, just FYI. @sorki, if you have time, would love to get your review too!

hnix-store-core/src/System/Nix/Internal/Hash.hs
```diff
@@ -0,0 +1,159 @@
...
+import qualified Crypto.Hash.SHA1      as SHA1
+import qualified Crypto.Hash.SHA256    as SHA256
+import qualified Data.ByteString       as BS
+import qualified Data.ByteString.Char8 as BSC
+import           Data.Bits             (xor)
+import qualified Data.ByteString       as BS
+import qualified Data.ByteString.Lazy  as BSL
+import qualified Data.Hashable         as DataHashable
+import           Data.List             (foldl')
+import           Data.Proxy            (Proxy(Proxy))
```
shlevy: Unnecessary import?

shlevy:
(I suppose there could possibly be some case where we want user code to have a `Digest a -> Digest (Truncated n a)` to avoid recalculating the hash, but that's a pure optimization that crosses conceptual boundaries)

imalsogreg:
I haven't been thinking of truncation as a type of hashing algorithm. In nix world, SHA256 is the only algo whose digest ever gets truncated, so I wonder if we should enforce that as: 

```
data HashAlgorithm =
  = MD5
  | SHA1
  | SHA256
  | TruncatedSHA256 Nat
``` 
?
Furthermore, `20` is the only truncation length ever used in nix. So, drop that type param too? Or should we leave it polymorphic in case _we_ ever want a larger family or algorithm tags?

Re: sometimes wanting a truncated digest and sometimes wanting an untruncated digest, here is an example. The top is part of some validity checking, where a FSO may be read and hashed without truncation. The bottom part is computing a store hash, where of course we _do_ truncate.

```
valid[p] ← "sha256:"+printHash32(hash_sha256(serialise(readPath(p))))
```
vs.

```
makePath(type, descr, name) =
nixStore+"/"+printHash32(truncate(20, hash_sha256(s))) +"-"+name
where
s = type+":sha256:"+descr+":"+nixStore+":"+name
```

shlevy:
> In nix world, SHA256 is the only algo whose digest ever gets truncated

`nix hash` can `--truncate` any hash, though it's true for the actual store interface we are only ever truncating sha256.

> Furthermore, 20 is the only truncation length ever used in nix. So, drop that type param too? Or should we leave it polymorphic in case we ever want a larger family or algorithm tags?

> If the type-level nats are too painful, we can specialize this to `Truncated160` for now.

:wink: 

> Re: sometimes wanting a truncated digest and sometimes wanting an untruncated digest, here is an example.

Let's look at the actual store path hashing code:

```C++
Path Store::makeStorePath(const string & type,
    const Hash & hash, const string & name) const
{
    /* e.g., "source:sha256:1abc...:/nix/store:foo.tar.gz" */
    string s = type + ":" + hash.to_string(Base16) + ":" + storeDir + ":" + name;

    checkStoreName(name);

    return storeDir + "/"
        + compressHash(hashString(htSHA256, s), 20).to_string(Base32, false)
        + "-" + name;
}
```

It's true here we're passing in a hash, and then computing a truncated hash. But *the truncated hash is not a truncation of the passed in hash*. Instead, the truncated hash is the truncated hash of a string which includes as one of its parts the hex-encoded representation of the input hash. The input hash can even be a completely different hash algorithm, e.g. if you have a fixed output derivation and provide a `sha512` instead of a `sha256`.

hnix-store-core/src/System/Nix/Internal/Hash.hs
imalsogreg: At the time it was leftover. Now it's needed for a use of `natVal`.

PR #31, opened 2019-03-10
Weird bit fiddling to make `printHashBytes32` work

Ericson2314:
> builtins.placeholder "asdf"

should be

> "/1x0ymrsy7yr7i9wdsqy9khmzc1yy7nvxw6rdp72yzn50285s67j5"

but was 

> "/x0ymrsy7yr7i9wdsqy9khmzc1yy7nvxw6rdp72yzn50285s67j5"

i.e. no leading 1

### DIFF
--- a/hnix-store-core/src/System/Nix/Internal/Hash.hs
+++ b/hnix-store-core/src/System/Nix/Internal/Hash.hs
@@ -125,7 +125,10 @@ newtype Digest (a :: HashAlgorithm) = Digest
 printHashBytes32 :: BS.ByteString -> T.Text
 printHashBytes32 c = T.pack $ concatMap char32 [nChar - 1, nChar - 2 .. 0]
   where
-    -- The base32 encoding is 8/5's as long as the base256 digest
+    -- The base32 encoding is 8/5's as long as the base256 digest.  This `+ 1`
+    -- `- 1` business is a bit odd, but has always been used in C++ since the
+    -- base32 truncation was added in was first added in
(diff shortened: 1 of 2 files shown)


hnix-store-core/src/System/Nix/Internal/Hash.hs
```diff
@@ -126,7 +126,7 @@ printHashBytes32 :: BS.ByteString -> T.Text
 printHashBytes32 c = T.pack $ concatMap char32 [nChar - 1, nChar - 2 .. 0]
   where
     -- The base32 encoding is 8/5's as long as the base256 digest
-    nChar = fromIntegral $ BS.length c * 8 `div` 5
+    nChar = fromIntegral $ ((BS.length c * 8 - 1) `div` 5) + 1
```
Ericson2314: See https://github.com/haskell-nix/hnix/commit/5b1117f6b90d547488dea21f7d1a9b85aa36aa02#diff-b5e9db944ee4f0a506a6978743b0752aR152

Ericson2314: https://github.com/NixOS/nix/blob/6024dc1d97212130c19d3ff5ce6b1d102837eee6/src/libutil/hash.hh#L70

Ericson2314: First added in d58a11e019813902b6c4547ca61a127938b2cc20 (year 2005).

PR #32, opened 2019-03-10
Implement readonly store path hashing

mightybyte:
This PR targets the `add-to-store` branch.

### DIFF
--- a/hnix-store-core/src/System/Nix/Internal/Hash.hs
+++ b/hnix-store-core/src/System/Nix/Internal/Hash.hs
@@ -12,25 +12,29 @@ Maintainer  : Greg Hale <imalsogreg@gmail.com>
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications    #-}
 {-# LANGUAGE TypeInType          #-}
+{-# LANGUAGE OverloadedStrings   #-}
 
 module System.Nix.Internal.Hash where
 
-import qualified Crypto.Hash.MD5       as MD5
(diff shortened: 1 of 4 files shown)


hnix-store-core/src/System/Nix/Internal/Hash.hs
```diff
@@ -80,8 +96,11 @@ hashLazy :: forall a.HasDigest a => BSL.ByteString -> Digest a
 hashLazy bsl =
   finalize $ foldl' (update @a) (initialize @a) (BSL.toChunks bsl)
 
+digestText32 :: forall a. HashAlgoText a => Digest a -> T.Text
+digestText32 d = algoString (Proxy :: Proxy a) <> ":" <> printAsBase32 d
```
imalsogreg: Is there a term in nix for "algo string with encoded hash"? The `Digest a -> Text` functions and similar become hard to keep track of, including the ones we've already got on `master`.

Some combination of (a) marking functions as internal in their Haddocks, (b) using more newtypes to limit which functions we can compose and (c) being strict about keeping functions in the right module may be our best defense here.

So `digestText32` and `digestText16` - are they more closely associated with Hashes in all their generality? Or do they have one specific use in `Path` or in `ReadOnlyStore`? Would it make sense to add a newtype here?

mightybyte: Yeah, I just did this because it was quick and obvious.  I could definitely see room for some newtypes that denote digests converted to a string with the hash algorithm prepended.

Issue #34, opened 2019-03-10
Add parser for base32 encoded byte strings


imalsogreg:
This should also be used to confirm whether or not the cargo-culted copy from the C++ implementation is actually needed for round-tripping or is just another example of bug-for-bug compatibility.

With this we can add several property tests, which should be made independent of the hashing tests.

sorki:
Part of #57

sorki:
This can be closed.

PR #35, opened 2019-03-10
Bit too late

sorki:
Lots of things mainly 
- path hashing + tests
- store now uses hashed paths
- tries to test itself against reference implementations (some of the tests started to fail, need to debug further)
- remote store accepts socket path and store path
- attempt to integrate with Nix.Derivation library

Sorry for getting this out late but I've got stuck last time mainly due to derivations and misunderstanding of paths - caused some work duplication but it should be merge-able.


sorki:
digestText16 can be implemented by means of `printAsBase16 (Digest bs) = printHashBytes16 bs` depending on whether we want to depend on base16-bytestring or not (performance might be better I guess).

I've also added `printHashAlgo` as it's required for building path strings.

hnix-store-remote/src/System/Nix/Store/Remote/Logger.hs
```diff
@@ -34,12 +37,23 @@ processOutput = go decoder
           case ctrl of
             e@(Error _ _) -> return [e]
             Last -> return [Last]
+            Read n -> do
```
sorki: StoreMonad now embeds Maybe Handle in its StateT so we can provide handle and use it to provide data when other side requests tunneling.

hnix-store-core/src/System/Nix/Internal/Hash.hs
```diff
@@ -76,6 +76,7 @@ class HasDigest (a :: HashAlgorithm) where
   initialize        :: AlgoCtx a
   update            :: AlgoCtx a -> BS.ByteString -> AlgoCtx a
   finalize          :: AlgoCtx a -> Digest a
+  hashName          :: T.Text
```
imalsogreg: This is a sensible place to put `hashName` (I think others have put this under its own typeclass, but simply using `HashDigest` seems better). `HasDigest` is not a great typeclass name under this change though. What's a name for this class that's neither (a) too specific for the methods it provides, (b) as general as `"Hash"`?

hnix-store-core/src/System/Nix/Internal/Hash.hs
```diff
@@ -96,41 +97,56 @@ hashLazy :: forall a.HasDigest a => BSL.ByteString -> Digest a
...
+-- | Convert any Digest to a base16-encoded string.
+printAsBase16 :: Digest a -> T.Text
+printAsBase16 (Digest bs) = printHashBytes16 bs
+
 -- | Convert any Digest to a base32-encoded string.
 --   This is not used in producing store path hashes
 printAsBase32 :: Digest a -> T.Text
 printAsBase32 (Digest bs) = printHashBytes32 bs
 
+-- | Print lowercased name of the hashing algorithm
```
imalsogreg: Can we remove this function? It perfectly overlaps with `hashName`.

hnix-store-core/src/System/Nix/Internal/Path.hs
```diff
@@ -0,0 +1,150 @@
...
+-- make output path from `PathName` digest and outputId which typically is "out"
+makeOutputPath :: (HasDigest a) => PathName -> Digest a -> T.Text -> Settings -> Text
+makeOutputPath pName digest outputId settings =
+  makeStorePath typ (adjustName pName) digest settings
+  where
+    typ = T.concat [ "output:", outputId ]
+    adjustName n | outputId == "out" = n
+    adjustName (PathName name) | otherwise = PathName $ T.concat [ name, T.pack "-", outputId ]
+
+type Recursive = Bool
```
imalsogreg: Nit: `data Recursive = Recursive | NotRecursive`  preferred over type alias.

hnix-store-core/src/System/Nix/Internal/Path.hs
```diff
@@ -0,0 +1,150 @@
...
+-- experimental
+-- Directory of the store
+type StoreDir = Text
+type Stored a = (StoreDir, a)
+
+-- wrap StoreDir and Path into tuple
+makeStored :: StoreDir -> Path -> Stored Path
+makeStored sl p = (sl, p)
+
+type PathType = Text
```
imalsogreg: Could we create a new datatype for this? I think it would be:
```
-- | Store object type is used when creating store paths, to differentiate
-- between paths for sources and derivation outputs (see thesis page 94)
data StoreObjectType =
    Source
  | Output Text -- ^ A build output (e.g. `Output "out"`)
```
or
```
-- | A single derivation build output (e.g. "out", "docs" etc) 
newtype BuildOutputName = BuildOutputName { getBuildOutputName :: Text }

-- | Store object type is used when creating store paths, to differentiate
-- between paths for sources and derivation outputs (see thesis page 94)
data StoreObjectType =
    Source
  | Output BuildOutputName -- ^ A build output (e.g. `Output (BuildOutputName "out")`)
```

hnix-store-core/src/System/Nix/Internal/Path.hs
```diff
@@ -0,0 +1,150 @@
...
+
+-- build <h> string which is a truncated base32 formatted SHA256 hash of <s>
+pathHash :: (HasDigest a) => PathType -> PathName -> Digest a -> StoreDir -> Digest PathHashAlgo
+pathHash typ pName digest storeLoc = System.Nix.Hash.hash . BSC.pack . T.unpack $
+  makePathDigestString typ pName digest storeLoc
+
+-- build <s> string which is hashed and used in makeStorePath
+-- <s> = "<pathType>:<hash_algo>:<base16_hash>:<storeDir>:<pathName>"
+-- (exposed for testing purposes only)
+makePathDigestString ::  (HasDigest a) => PathType -> PathName -> Digest a -> StoreDir -> Text
```
imalsogreg: If this function is only used in `makeStorePath`, can we move it into a `where` clause in `makeStorePath`? That could help to reduce the largish number of functions with similar-sounding names at the top level.

hnix-store-core/src/System/Nix/Internal/Path.hs
imalsogreg: Actually:
```
-- | Whether the hash is computedover the serialisation of the output path
-- or over the contents of the single non-executable regular file
-- See thesis page 107
data OutputHashMode = Recursive | Flat

**Comment by imalsogreg** [comment]
Lot of good stuff here, @sorki!

File: hnix-store-core/src/System/Nix/Internal/Hash.hs (hunk 2 above)
Comment (shlevy) [reply hnix-store-core/src/System/Nix/Internal/Hash.hs]: I disagree that this should be under `HasDigest` (which I've renamed `ValidAlgo` in my branch). Not every hash has a canonical name, e.g. the truncated ones don't (and shouldn't!).

File: hnix-store-core/src/System/Nix/Internal/Path.hs (hunk 4 above)
Comment (shlevy) [reply hnix-store-core/src/System/Nix/Internal/Path.hs]: This is still being broken up into manageable chunks,but see https://github.com/haskell-nix/hnix-store/commit/974a944bf5a30e1090acb8587817afa2433bd39a#diff-d6eea410e6c564641deccfa88ae99d1fR101

File: hnix-store-core/src/System/Nix/Internal/Path.hs [hunk 7]
```diff
@@ -0,0 +1,150 @@
...
+--
+-- XXX: should check for @PathHashAlgo length
+parsePath :: BSL.ByteString -> Maybe Path
+parsePath p = case name of
+    Nothing -> Nothing
+    Just n -> Just $ Path digest n
+  where
+  base = T.pack . takeBaseName . BSC.unpack . BSL.toStrict $ p
+  parts = T.breakOn "-" base
+  digest = Digest . BSC.pack . T.unpack . fst $ parts
```
sorki: I think this is incorrect and needs #34

sorki:
I would like to rebase and fix this at some point - probably after the Path changes are merged.

sorki:
Superseded by #59. All of the stuff from review is now addressed.

PR #36, opened 2019-03-10
Add makeFixedOutputPath

puffnfresh:


puffnfresh:
> Why do we special case this to SHA256?

At the moment `makeStorePath` is special cased to SHA256, which the function uses.

I've fixed the digest problem. Seems to be working.

shlevy:
Ah can you rebase on/merge in https://github.com/haskell-nix/hnix-store/pull/42 ?

hnix-store-core/src/System/Nix/ReadonlyStore.hs
```diff
@@ -28,5 +28,14 @@ makeTextPath nm h refs = makeStorePath ty h nm
   where
     ty = BS.intercalate ":" ("text" : map storePathToRawFilePath (HS.toList refs))
 
+makeFixedOutputPath :: (KnownStoreDir storeDir, ValidAlgo hashAlgo, NamedAlgo hashAlgo) => Bool -> Digest hashAlgo -> StorePathName -> StorePath storeDir
+makeFixedOutputPath recursive h nm =
+  makeStorePath ty h' nm
+  where
+    (ty, h') =
+      if recursive
```
shlevy: Only if recursive and sha256, right?

shlevy: https://github.com/NixOS/nix/blob/41ba5135e0a2d9940c4e8cbedff44eb1a47af782/src/libstore/store-api.cc#L200

puffnfresh: Ah maybe that's why I had it fixed to SHA256 previously. Will update!

puffnfresh: @shlevy I've updated, it will now only use "source" if recursive and SHA256.

Ericson2314:
We should add something about the references that are unstable in Nix, but this looks good to me.

PR #37, opened 2019-03-10
Initial work on S3 addToStore

puffnfresh:


hnix-store-s3/src/System/Nix/Store/S3.hs
```diff
@@ -0,0 +1,107 @@
...
+upsertFile
+  :: BucketName
+  -> Text
+  -> BS.ByteString
+  -> Text
+  -> AWS PutObjectResponse
+upsertFile bucketName pth d mimeType =
+  send (putObject bucketName (ObjectKey pth) (toBody d) & poContentType ?~ mimeType)
+
+data NarInfo
```
puffnfresh: Needs to go into core

puffnfresh:
Not tested at all. Of course!!!!

puffnfresh:
Just verified that this is now producing the exact same `.nar` and `.narinfo` files! :fireworks: 

```
$ diff -r hnix nix
Only in nix: nix-cache-info
```

I'm going to parameterise it by the `upsertFile`, so it's generalised to any binary store, then put it in `hnix-store-core`.

puffnfresh:
I'll also reuse `ValidPathInfo`

hnix-store-core/src/System/Nix/StorePathMetadata.hs
```diff
@@ -47,3 +52,42 @@ data StorePathTrust
...
+  { -- | URL to compressed Nar file.
+    narUrl :: Text
+  , -- | Form of compression used on Nar file.
+    narCompression :: Text
+  , -- | The hash of the nar serialization of the path.
+    fileHash :: !SomeNamedDigest
+  , -- | The size of the nar the path, in bytes.
+    fileBytes :: !(Maybe Word64)
+  }
+
```
sorki, line 67: I've completely missed this and recently wrote/extracted from Cachix a lib - https://github.com/sorki/nix-narinfo which is hopefully polymorphic enough to accommodate our types.

hnix-store-binary/src/System/Nix/Store/Binary.hs
```diff
@@ -0,0 +1,93 @@
...
+data BinaryStoreEffects m =
+  BinaryStoreEffects
+    { upsertFile :: Text -> BS.ByteString -> Text -> m ()
+    }
+
+narInfoFileFor
+  :: StorePath storeDir
+  -> Text
+narInfoFileFor p =
+  encodeBase32 (storePathHash p) <> ".narinfo"
```
sorki, line 39: This is now part of `-core`s `System.Nix.StorePath` called `storePathToNarInfo`.

hnix-store-core/src/System/Nix/Internal/Hash.hs
```diff
@@ -103,6 +103,16 @@ encodeBase32 (Digest bs) = Base32.encode bs
...
 
+-- | Encode a 'Digest' in algorithm:hash format.
+encodeDigest :: forall a. NamedAlgo a => Digest a -> Text
+encodeDigest d =
+  algoName @a <> ":" <> encodeBase16 d
+
+-- | Encode a 'Digest' in algorithm:hash format.
+encodeSomeDigest :: SomeNamedDigest -> Text
+encodeSomeDigest (SomeDigest d) =
+  encodeDigest d
```
sorki, line 114: This is pending via #64.

hnix-store-core/src/System/Nix/ReadonlyStore.hs
```diff
@@ -28,5 +29,14 @@ makeTextPath nm h refs = makeStorePath ty h nm
...
 
+makeFixedOutputPath :: (KnownStoreDir storeDir, ValidAlgo hashAlgo, NamedAlgo hashAlgo) => NarHashMode -> Digest hashAlgo -> StorePathName -> StorePath storeDir
+makeFixedOutputPath narHashMode h nm =
+  makeStorePath ty h' nm
+  where
+    (ty, h') =
+      case narHashMode of
+        Recursive -> ("source", h)
+        RegularFile -> ("output:out", hash ("fixed:out:" <> encodeUtf8 (encodeBase16 h) <> ":"))
+
```
sorki, line 40: This landed as part of #36.

hnix-store-binary/src/System/Nix/Store/Binary.hs
```diff
@@ -0,0 +1,93 @@
...
+
+addToStore
+  :: forall storeDir m. (MonadIO m, KnownStoreDir storeDir)
```
sorki, line 42: As of #59 (1f862114c67b7f1a203a1951d93c2cfa256f4257) `StorePath` is no longer parametrized by `KnownStoreDir storeDir`, store root `FilePath` is part of `storePathRoot` of `StorePath`.

```suggestion
  :: MonadIO m
```

hnix-store-binary/src/System/Nix/Store/Binary.hs
```diff
@@ -0,0 +1,93 @@
...
+  => StorePathName
+  -> FilePath
+  -> NarHashMode
+  -> BinaryStoreEffects m
+  -> m (StorePath storeDir)
```
sorki, line 47:
```suggestion
  -> m StorePath
```

sorki:
Looking good! Few things need to land and we can merge this. Your thoughts on effect systems and having GADT for ops like `addToStore`?

Ericson2314:
I would not bother with effect systems. Big waste of time.

sorki:
> I would not bother with effect systems. Big waste of time.

Care to elaborate on that?

I like how the effects compose compared to stack of transformers we use currently (mostly in `hnix` and `hnix-store-remote`) and came up with an `hnix-store-polysemy` experiment recently https://gist.github.com/sorki/10a035089e601045789357f37e7ab89c#file-a-hs-L32 that could be our high level effect, while parts of the store could be implemented in terms of effects again allowing for flexibility (e.g. running a store over TCP instead of socket, purely, local vs remote, s3..). We already have a home-brew effect system in both `hnix` and `hnix-store` so why not full blown effect system? (disregarding immaturity, which is gonna change when `eff` and related GHC proposal lands).

sorki:
I would like to salvage this, it bothers me it's sitting here for so long. Lots of pieces added here landed in core in some form.

Some random remarks, I won't add anymore inline comments as it makes the code hard to read:
- `NarHashMode` is no more and was replaced by [`ContentAddressMethod`](https://github.com/haskell-nix/hnix-store/blob/master/hnix-store-core/src/System/Nix/ContentAddress.hs#L28-L31). This is probably the least important thing since `addToStore` in Remote used to use a `Bool ^ -- recursive` which I've recently replaced with `newtype` but yesterday I found the same `Bool` in `ReadOnlyStore`, so I'll add that newtype to `-core` (So `NarHashMode` here will become `newtype Recursive = .. Bool`) - https://github.com/haskell-nix/hnix-store/issues/238

- It should use [nix-narinfo](https://github.com/sorki/nix-narinfo/), it has more fields but these are mostly `Maybe` or `Set` which could be left empty until required https://github.com/sorki/nix-narinfo/blob/master/src/Nix/NarInfo/Types.hs
  - `encodeContentAddressableAddress` used by `NarInfo` part is no more and was replaced by [`System.Nix.ContentAddress.buildContentAddress`](https://github.com/haskell-nix/hnix-store/blob/master/hnix-store-core/src/System/Nix/ContentAddress.hs#L69)
  - `encodeSomeDigest` was replaced by [`System.Nix.Hash.algoDigestBuilder`](https://github.com/haskell-nix/hnix-store/blob/master/hnix-store-core/src/System/Nix/Hash.hs#L210-L215) (`SomeDigest` and `SomeNamedDigest` are no more and they were replaced by `DSum HashAlgo Digest` or just `Digest`)

- I think the better name for `-binary` is `-local` since it seems local-store related (and `-binary` suggests that is has something to do with binary package / instances). I have a stub `-local` in one of my worktrees which is just this
```haskell
import qualified System.Linux.Mount
import qualified System.Linux.Namespaces

makeStoreWriteable :: IO ()
makeStoreWriteable = do
  System.Linux.Namespaces.unshare [System.Linux.Namespaces.Mount]
  System.Linux.Mount.rebind "/nix/store" []
```

It actually doesn't seem that hard, I might try cleaning up the PR after I'm done with #233 #235 #238

PR #39, opened 2019-03-21
Base32 tweaks

shlevy:


imalsogreg:
Nice! Also a good time to make char32 return a Char rather than a [Char]

imalsogreg:
@shlevy sorry for the late review! Was on mobile last night and my "submit review" button apparently didn't work.

PR #40, opened 2019-03-21
Migrate System.Nix.Path to System.Nix.StorePath

shlevy:
The new type acknowledges the store root, and is better named.


shlevy:
This is a draft, will mark ready once everything is migrated over or people want to start depending on this.

hnix-store-core/src/System/Nix/Internal/StorePath.hs
```diff
@@ -0,0 +1,109 @@
...
+
+-- | Smart constructor for 'StorePathName' that ensures the underlying
+-- content invariant is met.
+makeStorePathName :: Text -> Maybe StorePathName
+makeStorePathName n = case matchTest storePathNameRegex n of
+  True  -> Just $ StorePathName n
+  False -> Nothing
+
+-- | A type-level representation of the root directory of a Nix store.
+type StoreDir = Symbol
```
mightybyte: Does `StoreDir` really need to be at the type level?  I'm not seeing the value.  Why not keep it as a normal value-level field in `StorePath`?

hnix-store-core/src/System/Nix/Internal/StorePath.hs
```diff
@@ -0,0 +1,109 @@
...
+
+-- | A path in a Nix store.
+--
+-- @storeDir@: The root of the Nix store (e.g. \/nix\/store).
```
imalsogreg: Some text stolen from the thesis, for possible use in haddocks:
```
A store path is the full path of a store object. It has the following anatomy:
storeDir/hashPart-name
```
For `storePathHash` we could remind that this is the 160-bit hash digest, and that name is usually the package name - with the whole thing rendering to, for example `/nix/store/bwacc7a5c5n3qx37nz5drwcgd2lv89w6-hello-2.1.1`

hnix-store-core/src/System/Nix/Internal/StorePath.hs
```diff
@@ -0,0 +1,109 @@
...
+
+-- | Render a 'StorePath' as a 'ByteString'.
+--
+-- Would prefer to return a @RawFilePath@ here, but we want to avoid
+-- the dependency on the @unix@ package.
+storePathToByteString
+  :: forall storeDir . (KnownStoreDir storeDir)
+  => StorePath storeDir
+  -> ByteString
+storePathToByteString (StorePath {..}) = root <//> hashPart <//> name
```
imalsogreg: Should it be `root <//> hashPart <--> name`?

hnix-store-core/src/System/Nix/Internal/StorePath.hs
```diff
@@ -0,0 +1,109 @@
...
+  where
+    x <//> y = BS.concat [x, "/", y]
+    root = BC.pack $ symbolVal @storeDir Proxy
+    hashPart = encodeUtf8 $ encodeBase32 storePathHash
+    name = encodeUtf8 $ unStorePathName storePathName
+
+-- | Render a 'StorePath' as a 'FilePath'.
+--
+-- 'FilePath' is a terrible interface, but it's in base...
```
imalsogreg: Why do we need both this and `storePathToBytestring`?

hnix-store-core/src/System/Nix/Internal/StorePath.hs
```diff
@@ -0,0 +1,109 @@
...
+  :: forall storeDir . (KnownStoreDir storeDir)
+  => StorePath storeDir
+  -> FilePath
+storePathToFilePath (StorePath {..}) = root </> hashPart </> name
+  where
+    root = symbolVal @storeDir Proxy
+    hashPart = T.unpack $ encodeBase32 storePathHash
+    name = T.unpack $ unStorePathName storePathName
+
+-- | Render a 'StorePath' as 'Text'.
```
imalsogreg: Do we need this function? It seems like the user could instead use `decodeUTF8 . storePathToBytestring`. Then they could even choose whether they want to use the optimistic utf8 decoder or the safe one that runs in `Maybe`

hnix-store-core/src/System/Nix/Internal/StorePath.hs
```diff
@@ -0,0 +1,109 @@
...
+--
+-- Only sound when @storeDir@ is made up of valid unicode codepoints.
+-- Even then, using 'Text' to represent a file path is probably a bad
+-- idea... Prefer 'storePathToByteString' unless you *really* mean
+-- 'Text'.
+storePathToText
+  :: forall storeDir . (KnownStoreDir storeDir)
+  => StorePath storeDir
+  -> Text
+storePathToText (StorePath {..}) = root <//> hashPart <//> name
```
imalsogreg: root/hash-name

hnix-store-core/src/System/Nix/Internal/StorePath.hs
```diff
@@ -0,0 +1,109 @@
...
+    name = encodeUtf8 $ unStorePathName storePathName
+
+-- | Render a 'StorePath' as a 'FilePath'.
+--
+-- 'FilePath' is a terrible interface, but it's in base...
+storePathToFilePath
+  :: forall storeDir . (KnownStoreDir storeDir)
+  => StorePath storeDir
+  -> FilePath
+storePathToFilePath (StorePath {..}) = root </> hashPart </> name
```
imalsogreg: root/hash-name

hnix-store-core/src/System/Nix/Internal/StorePath.hs
shlevy: I'll update the comment with more details, but just a tidbit: There are 30 places in the C++ nix codebase where we have a runtime assertion that is only there because the type isn't strong enough :wink:

shlevy:
@imalsogreg Thanks for the feedback (and reminding me why I don't do late-night hacking), I'll update with your suggestions.

hnix-store-core/src/System/Nix/Internal/StorePath.hs
imalsogreg: Is there a way to take some arbitrary storeDir CLI argument and make a `KnownSymbol` from it?

shlevy: http://hackage.haskell.org/package/base-4.12.0.0/docs/GHC-TypeLits.html#v:someSymbolVal

shlevy: Added detailed docs explaining this decision.

hnix-store-core/src/System/Nix/Internal/StorePath.hs
shlevy: Fixed, :man_facepalming:

hnix-store-core/src/System/Nix/Internal/StorePath.hs
shlevy: We don't.

hnix-store-core/src/System/Nix/Internal/StorePath.hs
shlevy: Right, removed.

hnix-store-core/src/System/Nix/Internal/StorePath.hs
```diff
@@ -0,0 +1,141 @@
...
+-- | A path in a Nix store.
+--
+-- From the Nix thesis: A store path is the full path of a store
+-- object. It has the following anatomy: storeDir/hashPart-name.
+--
+-- @storeDir@: The root of the Nix store (e.g. \/nix\/store).
+--
+-- See the 'StoreDir' haddocks for details on why we represent this at
+-- the type level.
+data StorePath (storeDir :: StoreDir) = StorePath
```
imalsogreg: ghc 8.4.3 fails on this unless you use `TypeInType`

shlevy:
@imalsogreg Thanks for the review! Would be nice to set up CI here for all the GHC versions we care about, any chance you can take a swing at that?

shlevy:
@sorki Note the 4th commit here, we'll definitely bring back your work once it's ready!

shlevy:
@imalsogreg It seems like there's no way to bring a PR back into the "draft" stage :frowning:

shlevy:
This is ready for full review!

shlevy:
@imalsogreg Well, even once we actually have file putting/parsing doing the right thing, I'm still not sure we really want that exe... Wouldn't a repl module or a test be better? Or maybe even a proper repl cli tool?

imalsogreg:
@shlevy Yah, a repl or a test would be better. The exe was nice as an easy example to follow. But we still have the git history for that.

PR #41, opened 2019-03-23
Nar module cleanups

shlevy:


shlevy:
Still a draft, will mark ready once the module is up to spec

hnix-store-core/src/System/Nix/Internal/Nar.hs
```diff
@@ -0,0 +1,243 @@
...
+  , narCreateLink :: RawFilePath -> RawFilePath -> m ()
+  , narIsExec     :: RawFilePath -> m IsExecutable
+  , narIsDir      :: RawFilePath -> m Bool
+  , narIsSymLink  :: RawFilePath -> m Bool
+  , narFileSize   :: RawFilePath -> m Int64
+  , narReadLink   :: RawFilePath -> m RawFilePath
+}
+
+
+-- Directly taken from Eelco thesis
```
imalsogreg: Do we know what this comment was referring to? The whole module, Nar effects? Probably not the Nar newtype.

hnix-store-core/src/System/Nix/Internal/Nar.hs
```diff
@@ -0,0 +1,243 @@
...
+        parens m = str "(" >> m >> str ")"
+
+        -- Do not use this for file contents
+        str :: BSL.ByteString -> B.Put
+        str t = let len = BSL.length t
+            in int len <> pad len t
+
+        putContents :: Int64 -> BSL.ByteString -> B.Put
+        putContents fSize bs = str "contents" <> int fSize <> (pad fSize bs)
+        -- putContents fSize bs = str "contents" <> int (BSL.length bs) <> (pad fSize bs)
```
imalsogreg: Stale comment

imalsogreg (changes requested):
Good reorg. Only nits to change.

imalsogreg:
(meta). Does `Draft` mean, hold off on commenting for now?

shlevy:
Draft means I'm still expecting more to come (in particular, so far I've only touched NarEffects and still need to go through the rest and add haddocks etc.), but still ready for comments on what's come so far.

Also, I'm trying to make to so every commit is in a workable state, so that if someone *wants* to depend on this before I'm done we can merge in and I can continue the rest later.

sorki:
The last commit [e5f1c7d](https://github.com/haskell-nix/hnix-store/pull/41/commits/e5f1c7d9bcf618c3d1351fb006699285402f42d5) was ported to `hnix-store-nar` in #259 (prompted via #252). 

The rest is kind-of obsolete maybe except for the `RawFilePath` patch but the new `filepath` would be a better fit, see also https://github.com/haskell-nix/hnix-store/issues/237

Closing

PR #45, opened 2019-03-28
hnix-store-remote README: fix example

4z3:
Minor changes to the example so it actually works.


Ericson2314:
Opened https://github.com/haskell-nix/hnix-store/issues/47 to track making this sort of thing impossible ;).

PR #46, opened 2019-03-29
Add release file for hydra

imalsogreg:
I copied most of this from hnix. @ryantrinkle, would you mind adding this project to your hydra instance, and we can use that for CI?

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -36,9 +36,10 @@ let
       then pkgs.lib.composeExtensions orig.overrides overlay
       else overlay;
   };
-  
+
 in {
   haskellPackages =
     pkgs.haskell.packages.${compiler}.override overrideHaskellPackages;
   inherit pkgs;
+  nixpkgs = pkgs;
 }
--- a/default.nix
+++ b/default.nix
@@ -1,5 +1,34 @@
-{ pkgs ? import <nixpkgs> {} }: let
-  overlay = import ./overlay.nix;
+{
+  compiler ? "ghc863"
+, nixpkgs-rev ? "120eab94e0981758a1c928ff81229cd802053158"
+, nixpkgs-sha256 ? "0qk6k8gxx5xlkyg05dljywj5wx5fvrc3dzp4v2h6ab83b7zwg813"
+
+, pkgs ? import (builtins.fetchTarball {
+    url = "https://github.com/NixOS/nixpkgs/archive/${nixpkgs-rev}.tar.gz";
+    sha256 = nixpkgs-sha256;
+  }) {
+    config.allowUnfree = true;
+    config.allowBroken = false;
+  }
+
+, doStrict  ? false
+, doTracing ? false
+
+}:
+
+let
+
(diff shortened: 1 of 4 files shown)


release.nix
```diff
@@ -0,0 +1,23 @@
...
+
+      # Broken
+      # { compiler = "ghc802"; doStrict = false; doTracing = false; }
+      # { compiler = "ghc802"; doStrict = false; doTracing = true;  }
+
+      # Deprecated
+      # { compiler = "ghc822"; doStrict = true;  doTracing = false; }
+      # { compiler = "ghc822"; doStrict = true;  doTracing = true;  }
+
+      # Broken
```
shlevy: Why? Is it our code or nixpkgs?

imalsogreg:
@shlevy Right you are. And jobsets.nix looks necessary too. Added.

release.nix
imalsogreg: I haven't worked it out yet.
```
[greghale@cedric:~/code/hnix-store]$ nix build -f default.nix haskellPackages.hnix-store-core --argstr compiler ghcjs
error: attribute 'override' missing, at /home/greghale/code/hnix-store/default.nix:34:7
(use '--show-trace' to show detailed location information)
```

default.nix
```diff
@@ -1,16 +1,45 @@
...
+  }
+
+, doStrict  ? false
+, doTracing ? false
+
+}:
+
+let
+
+  overlay = self: super: builtins.mapAttrs (k: v: (
```
shlevy: Rather than doing mapAttrs here maybe we should generate both this and the underlying overlay from a list of package names?

default.nix
```diff
@@ -1,16 +1,45 @@
...
       then pkgs.lib.composeExtensions orig.overrides overlay
       else overlay;
   };
+
 in {
   haskellPackages =
-    pkgs.haskellPackages.override overrideHaskellPackages;
-  haskell844Packages =
-    pkgs.haskell.packages.ghc844.override overrideHaskellPackages;
+    pkgs.haskell.packages.${compiler}.override overrideHaskellPackages;
```
shlevy: This is where the ghcjs bit is failing. Unclear why it's not a normal haskell package set though.

hydra.json
```diff
@@ -0,0 +1,29 @@
+{
+  "enabled": 1,
+  "hidden": true,
```
shlevy: Why?

sorki:
This is very dated and probably not that important either unless someone wants to run Hydra for `haskell-nix`

https://github.com/haskell-nix/hnix-store/commit/994a3ae01ff6e6ca5eeb346e5244a6dff8c1578d added `matrix.nix` which uses [`ci.dhall`](https://github.com/haskell-nix/hnix-store/blob/a056181d7ea377899ba028de41aad206ce7ea6c4/.github/workflows/ci.dhall) as a single source of truth for a list of supported GHCs - that could be used to populate `release.nix` in future in similar manner.

PR #48, opened 2019-04-03
StorePath: Derive Eq to make StorePathSet usable.

shlevy:


shlevy:
It can, though it shouldn't be needed for this case since we're using `HashSet`

puffnfresh:
@shlevy yes that's right. Not needed for this exact case. Just in general.

shlevy:
https://github.com/haskell-nix/hnix-store/commit/671d3c58dd485ba0372d91891498180feaa8bfa5

PR #50, opened 2019-06-25
Add computeFSClosure

puffnfresh:
From some quick tests, this seems to compute the same results.

### DIFF
--- a/hnix-store-remote/src/System/Nix/Store/Remote.hs
+++ b/hnix-store-remote/src/System/Nix/Store/Remote.hs
@@ -10,10 +10,13 @@ module System.Nix.Store.Remote (
   , syncWithGC
   , optimiseStore
   , verifyStore
+  , computeFSClosure
   ) where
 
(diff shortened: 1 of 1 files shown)


hnix-store-remote/src/System/Nix/Store/Remote.hs
```diff
@@ -32,3 +35,25 @@ verifyStore :: CheckFlag -> RepairFlag -> MonadStore Bool
...
   putBool check
   putBool repair
+
+computeClosure
+  :: (Monad m, Foldable f, Monoid a)
+  => (a -> m (f r))
+  -> (r -> a)
+  -> a
+  -> m a
+computeClosure lookupReferences findReferences = go
```
puffnfresh: It's not awful but I feel like there might be something I'm missing to simplify this.

puffnfresh:
`referencesVP` got deleted so I'll have to update this.

Issue #54, opened 2019-12-31
GHC 8.8.1


domenkozar:
regex-tdfa-text needs a bump (or need to pick a different regex library ).

domenkozar:
Ah it's deprecated: `Deprecated. in favor of regex-tdfa`

dhess:
@domenkozar Easy peasy, see #55 

Note that you'll need to build `hnix` from git, as the most recent Hackage version also has a `regex-tdfa-text` dependency.

domenkozar:
Thanks :)

PR #55, opened 2020-01-26
regex-tdfa 1.3.1.0 replaces regex-tdfa-text.

dhess:
See https://github.com/haskell-hvr/regex-tdfa/issues/4


dhess:
Fixes #54.

Infinisil:
Can this be in a new hackage release so nixpkgs' build can be fixed? Or alternatively just a new cabal file revision

PR #58, opened 2020-03-05
fix nar test lookup for src directory

sorki:
Now also looks in `hnix-store-core/src` as well and won't fail
even if either of them doesn't exist (which is not a big deal
in this case).


imalsogreg:
Thanks @sorki ! Let's use this for now. Down the road replace this test with one that doesn't use the library's source at all. I think I added that test as a quick example for myself of a realistic directory, but it feels weird to be using the source code as test-data like this in the long term.

Would you mind adding a TODO comment to you PR reminding us to come back to this?

PR #63, opened 2020-04-05
hnix-store-remote: Remove ExceptT, use DList to accumulate log.

TravisWhitaker:
This PR refactors two bits in the worker protocol implementation:

- A superfluous ExceptT layer is removed. IO is at the base of the stack anyway, so nothing is gained from adding and additional, out-of-band error handling mechanism; callers that care about IO failure will have a set of exception handlers anyway.
- DList is used to accumulate the log. It seems that for long-running interactions with the store there should be a way to stream these log chunks, but this is a partial improvement that isn't opinionated about how the streaming is done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 dist
 dist-newstyle
 .ghc.environment*
+cabal.project.local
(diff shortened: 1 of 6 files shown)


sorki:
Thank you, your PR is similar to #52 which uses pipes. We need to get #59 merged and then we can decide on streaming and further improvements.

Anton-Latukha:
Added a `DList` benchmarks into: https://github.com/haskell-perf/sequences

It shows that `DList` actually really is a solution in itself. `O(1)` cons, snoc, append - laziness does the streaming.

PR #70, opened 2020-07-27
Database

sorki:
PoC of database compatible with Nix ones at `/nix/var/nix/db/db.sqlite` (schema version 10). 

Allows us to test roundtrips of content addressable addresses and store paths using systems db. 

- read only operations implemented
- can't create indexes
- `ValidPath` type is the same as the one in `-core`, glue conversion code is missing.

Test with 
```
nix-shell
cabal run hnix-store-db

# In repl
> test
> bench
```

This runs queries vs your local database (owned by root so it's safe to play with this), for me currently results in

```
*** Exception: PersistMarshalError "*** Exception: ("string","fixed:sha512:2dc6n9ggmbjq8z6ahkcs3bk6yb4z7c56b19fyrnzw9a2ykpjsbllcvpc0cnb279h8s9pl42l6by709bsjg9g7v527k5j29zppamf73b")
```

because we don't have base64 yet.

cc @layus

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,10 +1,14 @@
 packages:
   ./hnix-store-core/hnix-store-core.cabal
+  ./hnix-store-db/hnix-store-db.cabal
   ./hnix-store-remote/hnix-store-remote.cabal
 
 -- till https://github.com/obsidiansystems/dependent-sum/pull/80
 allow-newer:
   dependent-sum:some
+
+package hnix-store-db
+  flags: +build-readme +build-bench
 
 package hnix-store-remote
   flags: +build-readme +io-testsuite
--- a/cabal.project.local.ci
+++ b/cabal.project.local.ci
@@ -3,5 +3,8 @@ tests: True
 package hnix-store-core
   ghc-options: -Wunused-packages -Wall -Werror
 
+package hnix-store-db
+  ghc-options: -Wunused-packages -Wall -Werror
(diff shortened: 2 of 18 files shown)


sorki:
Last two commits add `base64` support and SHA512 seems to be handled correctly now.

layus:
Nice job :-) Will try to have a look at it in the near future

domenkozar:
https://github.com/cachix/cachix/blob/master/cachix/src/Cachix/Client/Store.hs#L42 how to properly call sqlite.db across the platforms.

layus:
That's a lot of `fixups !` ;-).
Do you need me to test it ?

sorki:
> That's a lot of `fixups !` ;-). Do you need me to test it ?

Ohai! Soon! It needs #231 (that's why it is so big) and couple more type annotations / transfomers :(( but it works well (at least on GHC947 hehe)

sorki:
> https://github.com/cachix/cachix/blob/master/cachix/src/Cachix/Client/Store.hs#L42 how to properly call sqlite.db across the platforms.

The link is dead but I've checked the history - all that's required is `file:` prefix? I've added that as it seems better to have it explicit.

sorki:
> That's a lot of `fixups !` ;-). Do you need me to test it ?

Ok, yes! Should be good now.

PR #61, opened 2020-03-12
Stream to and from NAR format

imalsogreg:
Addresses #51 

`System.Nix.Nar` changes its API - the main functions become roughly (not exactly)
  - `buildNarIO :: FilePath -> Handle -> IO ()` Create a NAR from a regular filesystem object, stream it out on the Handle.
  - `unpackNarIO :: Handle -> FilePath -> IO ()` Recreate filesystem object from a NAR file accessed by the Handle

The types `Nar` and `FileSystemObject` are no longer parts of the external API, although they still exist and help the implementation somewhat. Removing `FileSystemObject` was necessary to let NAR construction/consumption work in constant space, since parsing one to completion implies holding it all in memory. In the future we could bring back this return type by stripping out the files themselves. And it would certainly be useful to have that as a parse result, so that `hnix-store` could analyze the spine of the Nar.

Test tests all pass with `+RTS -M10m` (limiting heap size to 10Mb), and several of them produce NARs larger than 100Mb. So, we have streaming. There is also a test that packs/unpacks a directory with 1000 files into a NAR and ensures that we do not hold onto file descriptors for longer than needed (as #51 discovered for `master`).

PR #71, opened 2020-07-27
Vigorous io update

sorki:
PR against vigorous-io branch (#61) - buildable again, bounds, dropped dontCheck. 

cc @imalsogreg


imalsogreg:
Thanks, @sorki ! There is one failing test in alga when I build your branch:
```

sorki:
Dropped the dontCheck removal commit. It builds fine for me with recent nixpkgs and ghc 8.10 so I've removed it. It took a long time though.

imalsogreg:
@domenkozar I think this will fix the error you were seeing.

PR #72, opened 2020-11-01
WIP: Lift the monad stack to a monad stack transformer

layus:


hnix-store-remote/src/System/Nix/Store/Remote.hs
```diff
@@ -210,7 +214,7 @@ buildDerivation p drv buildMode = do
     -- XXX: reason for this is unknown
     -- but without it protocol just hangs waiting for
     -- more data. Needs investigation
-    putInt 0
+    putInt (0 :: Int)
```
sorki, line 217: I was keeping this one pop up in warnings intentionally as it's not clear why we need that.

layus: Yep, good catch. this is completely unrelated.

layus:
At this point everything works well. But hell, it made the code sooooo slow. `--eval -E '"${(import /home/gmaudoux/projets/nixpkgs { }).hello}"'` is now at 50s (from 22 before). And that code does not even use nar streaming, which is the very reason why the code is so complicated.

The obvious solution is to use a handler. But we may also get good performance with effect systems, where all the hype is ATM :-).

Unless someone has good insights on how to use mtl stacks correctly. My intuition tells me that there is something very wrong about the implem in this PR.

@bqv Fun stuff is also happening here, and in the sibling hnix PR https://github.com/haskell-nix/hnix/pull/755

PR #64, opened 2020-05-10
Improve digests generation and parsing

layus:
This pr is based on top of #59 and fixes the only two isuues I encountered while writing prim_derivationStrict (place here the pr number in hnix when created).

Only the last two commits are relevant here.

https://github.com/haskell-nix/hnix-store/commit/0099f6d0cb3f75ed492cd2e4546fc334fe98e3b5 has a proper `Text -> SomeNamedDigest` function. Thechnically base64 hashes are not handled, but I did not enouter them in my tests (builing the `hello` package)

 https://github.com/haskell-nix/hnix-store/commit/488fe78dce41a829bf0aa7436d3fd6f30771f772 fixes some quirks in hashes computations discovered by patiently comparing hnix and nix-instantiate outputs (I learned even more about dwarffs and gdb in the process :-))

PR #73, opened 2020-11-03
Sha512 and base64 support

layus:
Based on top of #64 which is fairly trivial and contains only small fixes.


sorki (approved):
LGTM, tested with my systems Nix database. I'll try to fix `contentAddressableAddressBuilder` in similar manner and we can possibly add some round-trip tests for it.

PR #74, opened 2020-11-10
Fix text encoding to utf8

layus:
For non ascii characters, the custom functions to convert text to/from bytestrings do not work as intended. They mangle the strings sent to the store, resulting in incorrect string content added to the store.
Found the bug in the serialisation of a .drv file.


sorki (approved):
LGTM. I might refactor this to use full qualified names instead of these Utils at some point.

PR #75, opened 2020-11-11
[Quickfixes] Fix tests and build of hnix-store-remote

layus:
- Propagate changes induced by the nar streaming code to the remote store package
- Fix a test quirk with queryPathInfoUncached


sorki (approved):
Thank you. Note to self - add changelog entry due to removal of `addToStoreNar`.

PR #93, opened 2020-12-17
Hash module refactor

Anton-Latukha:
Making the code more refactorable.

### DIFF
--- a/hnix-store-core/src/System/Nix/Internal/Hash.hs
+++ b/hnix-store-core/src/System/Nix/Internal/Hash.hs
@@ -38,7 +38,7 @@ import           Data.Coerce            (coerce)
 data BaseEncoding
   = Base16
   | Base32
-  -- | ^ Nix has a special map of Base32 encoding
+  -- ^ Nix has a special map of Base32 encoding
   | Base64
 
 -- | The universe of supported hash algorithms.
@@ -53,6 +53,22 @@ data HashAlgorithm
     -- ^ The hash algorithm obtained by truncating the result of the
     -- input 'HashAlgorithm' to the given number of bytes. See
     -- 'truncateDigest' for a description of the truncation algorithm.
+
+class HashProperties a
+ where
+  canonicalHashName :: a -> Text
+  canonicalHashLen :: a -> Int
+
+instance HashProperties HashAlgorithm
+ where
+  canonicalHashName SHA256 = "sha256"  -- SHA256 is the most used in Nix - so it matches first
+  canonicalHashName MD5    = "md5"
+  canonicalHashName SHA1   = "sha1"
+  canonicalHashName SHA512 = "sha512"
+  canonicalHashLen  SHA256 = 32
+  canonicalHashLen  MD5    = 16
+  canonicalHashLen  SHA1   = 20
+  canonicalHashLen  SHA512 = 64
 
 -- | The result of running a 'HashAlgorithm'.
 newtype Digest (a :: HashAlgorithm) =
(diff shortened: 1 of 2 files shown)


Anton-Latukha:
BTW what `;` in the `canonicalLenIf` meant, it seems just a desugared vacuous `;`, maybe someone programmed in some C++ right before it 8).

Anton-Latukha:
This seems to be the extent to where the refactor of `Hash.hs` can go without changing the code/type level logic.

After this point the TypeApplication and DataKind wall arrives. And touching it requires refactoring type level, interface, connectors through the projects. Well, at that point - if it further requires a ripping-out and rebuilding the whole Hash system - that is a type to introduce the new library and have a new interface.

After the refactor settles-in, it further may move functional Haskell code into TypeApplication and DataKinds if that brings that big of a boost in the performance.

But currently - that is as much classic refactor as possible in Hash before the major shift.

Anton-Latukha:
Well, maybe there is a middle-PR where the current Hash system moves into classical λ(2)ω.

hnix-store-core/src/System/Nix/Internal/Hash.hs
```diff
@@ -53,6 +53,22 @@ data HashAlgorithm
...
+instance HashProperties HashAlgorithm
+ where
+  canonicalHashName SHA256 = "sha256"  -- SHA256 is the most used in Nix - so it matches first
+  canonicalHashName MD5    = "md5"
+  canonicalHashName SHA1   = "sha1"
+  canonicalHashName SHA512 = "sha512"
+  canonicalHashLen  SHA256 = 32
+  canonicalHashLen  MD5    = 16
+  canonicalHashLen  SHA1   = 20
+  canonicalHashLen  SHA512 = 64
```
layus, line 70: This makes the function partial because you cannot define it for Truncated. Requiring a Named algo does not work either.

We could split Truncated outside of HashAlgorithm type. That would also prevent the weird possibility to create a `Truncated (Truncated hash)`

layus (approved):
Funny that you must make a refactor to make future refactors easy :-D

hnix-store-core/src/System/Nix/Internal/Hash.hs
Ericson2314: In C++ nix i swear the only point of truncated was the intermediate result when calculating store paths. I wonder if we can get rid of it entirely in both code bases

Anton-Latukha: Idea is that to move code closer&closer to Cryptonite interface. Library has this `canonicalHash{Name,Len}` info in instances.

Put truncated separately seems to be the minimal core compromise, it allows/would allow to use hash libraries directly and see where Truncated is really used.

Slipped the idea of SRI hash form to Cryptonite. Since that hash form is minimal.

The truncated hash form is a more rare broad sensitive question, do not think it should or would or be included in lib.

My plan is to take a response about SRI in Cryptonite, ask recommendations to extend the library, and then maybe for the SRI and Truncated hash create separate tiny packages. since they are legitimate but rarely used cases, there can be some other users for them.

layus: Oh, silly me. Truncated has a canonicalHashLen. It's just the truncation length :-)

Anton-Latukha: It would, but that does not matter to the software design, redefining the core `HashAlgorithm` type means redefining the whole library interface, so I thought and after our back&forth I ensured that Trunkated should be given separately.

Anton-Latukha:
In the module and use of TypeAplication with promotion into DataKinds the code is hard to replace, refactor. DataKinds is λp field feature btw, seems like type application makes it even harder to reason/work with it. At least for me currently.

And code needs to be refactored into an elegant form before lifting algorithms into types and enforcing type-level performance tuning into it.

Overall looked further and seen that due to `Internal.Hash` does full interface implementation and its connection with introduced type classes and type-level programming &TypeAplication with promotion DataKinds, almost all module gets replaced by `cryptonite` and really what is left - is `base` codec functions, and I am puzzled what `base` codec functions do in Hash module and why they specialized particularly on Digest types and nothing else. Since Nix has custom Base32, the Base interface needs to be formed, in a Base module.

Trunkated code deserves to be in separate module.

Lets do a normal refactor on the library interface and then look what we need further and how to hook it up.

I started a main refactor.

Anton-Latukha:
I use this PR code as reference, and rebased work on it. So if anything survives the refactor - results of the PR would be in the resulting code.

PR #94, opened 2020-12-19
Tmp override fix of cryptohash-sha512

### LINKED ISSUE #90: Please, fix cryptohash-sha512 temporarily
> Neat. I think you can try using cabal.project to add cryptohash-sha512 PR haskell-hvr/cryptohash-sha512#5 temporarily (https://cabal.readthedocs.io/en/3.4/cabal-project.html#specifying-packages-from-remote-version-control-locations).
> 
> `sorki`: https://github.com/haskell-nix/hnix-store/pull/88#issuecomment-744534070

Anton-Latukha:
`cryptohash-sha512` unmaintained, PR adds support of `base` 4.14 (GHC 8.10) through `cabal.project`

Closes #90.

### DIFF
--- a/hnix-store-core/cabal.project
+++ b/hnix-store-core/cabal.project
@@ -1,0 +1,6 @@
+packages: .
+
+source-repository-package
+    type: git
+    location: https://github.com/Anton-Latukha/cryptohash-sha512
+    tag: 82ffc68fdfc1ce3745368cd0ddc22cab5a76c769
--- a/hnix-store-remote/cabal.project
+++ b/hnix-store-remote/cabal.project
@@ -1,0 +1,6 @@
+packages: .
+
+source-repository-package
+    type: git
+    location: https://github.com/Anton-Latukha/cryptohash-sha512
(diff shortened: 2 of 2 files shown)


Anton-Latukha:
Recently I migrated Core to the `base16-bytestring` >= 1.

But because Remote uses Core from Hackage - it left using `base16-bytestring` < 1.

To support 8.10 in Cryptohash-sha512 - there was a need to forse some `base16-bytestring`

So...

In the Nix the projects builds, because they use local builds of one-another.

Cabal builds use Hackage. And there is no way to solve the Cabal build, because if in monorepo to force local source - Hackage upload&build would fail.

So, this seems the right state of affairs.

Remote needs a Core release to work.

Anton-Latukha:
Remote fails because of wrong bounds on the Hackage HNix-Core, since it allows but does not support `base16-bytestring >= 1`.

Recently I migrated Core to `base16-bytestring >= 1` - but that is in `master` - Core release would fix the Remote.

Issue #67, opened 2020-06-16
Lets have a CI, so it builds and tests stuff


Anton-Latukha:
In HNix I banged-out some pretty portable setup for Haskell:

it consists of:
1. https://github.com/haskell-nix/hnix/blob/master/default.nix which exposes `Nixpkgs haskell.lib` API.
2. https://github.com/haskell-nix/hnix/blob/master/build.sh which exposes `Nixpkgs haskell.lib` API.
3. https://github.com/haskell-nix/hnix/blob/master/.travis.yml which exposes .., you know the shtick ...

That resulted in: https://travis-ci.org/github/haskell-nix/hnix

Which helped to have and keep proper `cabal sdist`.
And strict build on GHC 8.6.5 currently. So no no warnings in PRs.
Also Hackage Matrix: https://matrix.hackage.haskell.org/#/package/hnix

CI even does GHCJS build with Cachix.

I and Nix already made fell on all rakes we had, so everything is ironed-out now.

---

What do you think?
We can easily migrate this CI setup to GitHub Actions, I would be up for it.

sorki:
Thank you for all the work on this! 

I'm weakly against switching to GitHub Actions as it feels kind-of unfair to Travis and further locks projects on GitHub. 

I'm quite biased on this though as I'm working on a CI built on top of `nix/hnix` that is hosting platform agnostic and most importantly allows self-hosting. It's not quite ready yet - I'll add a proposal to https://github.com/haskell-nix/hnix/issues/593 when I manage to iron out last few issues.

Anton-Latukha:
"Unfairness" to Travis I do not feel anymore.

Travis CI is a project of a proprietary company, proprietary company owners sold everything to the cynical firm, which right-away fired almost all workers including majors of Travis CI project.

Be "loyal" to Travis CI at this point it is like to be loyal to the empty proprietary brand name at this point. Like believing the lies that real Atari exists.

If we would compare Travis CI and GitHub CI now - there is no difference, both a proprietary closed software services, with the difference that Travis CI owners do not cared/care about Travis CI future at this point, while Microsoft, in fact, has a much better reputation now and cares about GitHub and products and has much more lavish abilities and generosity.

[...]

PR #97, opened 2020-12-22
CI: GitHub: Tmp allow broken/try to build broken Nix closures

Anton-Latukha:
SInce projects are broken quite frequently - introducing tmp `allowBroken` fix.
Works towards #67.


Anton-Latukha:
Now for Remote the nix-build ignores the Nixpkgs `broken` attribute, and we received a couple of issues that need to be addressed in the Remote.

PR #102, opened 2020-12-30
allowBroken cryptohash; base16-bytestring dep epochs support

Anton-Latukha:
* `allowBroken cryptohash` allows to check Nix dev env builds.

* Reintroduced tests to support the old `base16-bytestring` epoch, since it is trivial.


sorki:
Thanks!

Anton-Latukha:
Yep, sorry for this self-merges, cleaning-up the state here after myself also. So it is understandable what is the state of affairs. Didn't knew when you would show up, so finished the clean-up.

So now the state of affairs is seen.

Issue #101, opened 2020-12-30
Please, solve the Cabal build dependency design issue between the projects


Anton-Latukha:
`hnix-store-core` & `hnix-store-remote` separately `cabal v2-build` fine.

The HLS founds the cradle in `hnix-store` home directory, as it should - because `cabal.project`:
```
packages: ./hnix-store-core/*.cabal ./hnix-store-remote/*.cabal
```
Is put there.

This directly ties the two project dependencies together - and now HLS does not work:
```
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: hnix-store-core-0.3.0.1 (user goal)
[__1] trying: cryptohash-sha512-0.11.100.1 (dependency of hnix-store-core)
[__2] next goal: base (dependency of hnix-store-core)
[__2] rejecting: base-4.14.1.0/installed-4.14.1.0 (conflict: cryptohash-sha512
=> base>=4.5 && <4.14)
[__2] skipping: base-4.14.1.0, base-4.14.0.0 (has the same characteristics
that caused the previous version to fail: excluded by constraint '>=4.5 &&
<4.14' from 'cryptohash-sha512')
[__2] rejecting: base-4.13.0.0, base-4.12.0.0, base-4.11.1.0, base-4.11.0.0,
base-4.10.1.0, base-4.10.0.0, base-4.9.1.0, base-4.9.0.0, base-4.8.2.0,
base-4.8.1.0, base-4.8.0.0, base-4.7.0.2, base-4.7.0.1, base-4.7.0.0,
base-4.6.0.1, base-4.6.0.0, base-4.5.1.0, base-4.5.0.0, base-4.4.1.0,
base-4.4.0.0, base-4.3.1.0, base-4.3.0.0, base-4.2.0.2, base-4.2.0.1,
base-4.2.0.0, base-4.1.0.0, base-4.0.0.0, base-3.0.3.2, base-3.0.3.1
(constraint from non-upgradeable package requires installed instance)
[__2] fail (backjumping, conflict set: base, cryptohash-sha512,
hnix-store-core)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: base, hnix-store-core,
cryptohash-sha512
```

If one works on the dependency migration in the `hnix-store-core` - IDE stops to work, which is convenient.

Tying the projects with `cabal.project` this way would always complicate more than solve because it goes against the design of the Haskell ecosystem.

---

[...]

PR #103, opened 2020-12-30
Release 0.4

sorki:
Rendered core ChangeLog https://github.com/sorki/hnix-store/blob/release0.4/ChangeLog.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,10 @@
-hnix-store
-===========
+# hnix-store
 
 A Haskell interface to the [Nix] store.
 
 [Nix]: https://nixos.org/nix
 
-Rationale
------------
+## Rationale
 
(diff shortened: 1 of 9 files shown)


Anton-Latukha:
Looks good. But very strange.

sorki:
I would rather not make this a multi-stage process and instead release both at once when there changes to both libs.

Anton-Latukha:
Well, I just could be:
1. completely silent;
2. say something and be completely honest.

Anton-Latukha:
Thanks for doing this. The old stale situation was bugging me out for several days.

sorki:
Well be honest, I see your reasoning about this in #101 and I don't completely agree (will respond later). Provided I use a script to upload both + docs, `core` goes first and `remote` lands right after claiming support for the released core.

PR #110, opened 2021-01-09
cabal: rm cryptohash-sha512 override

Anton-Latukha:
https://github.com/haskell-hvr/cryptohash-sha512/pull/5#issuecomment-757068529

Nixpkgs is Ok with all this, it does not import these overrides and have own separate mechanics for the overrides. Since cryptohash-sha512 package is essentially in the unmaintained state - I think it is not worth cleaning the Nixpkgs description because soon (on soon arriving 9.0) - it may be needed again in Nixpgs store, and to that time I hope we make a transiiton.

### DIFF
--- a/hnix-store-core/cabal.project
+++ b/hnix-store-core/cabal.project
@@ -1,6 +1,1 @@
 packages: .
-
-source-repository-package
-    type: git
-    location: https://github.com/Anton-Latukha/cryptohash-sha512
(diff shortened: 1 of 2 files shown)


Anton-Latukha:
This cleaning has a nice side-effect that HNix now does not need to recompile the cryptohash and Store packages every time.

PR #115, opened 2021-01-14
Cleanup/warnings

sorki:
All of them except one, which has a comment.


Anton-Latukha:
Parsed through. Good changes.

Everything seems safe.

The moment where DataKind promotions happen I can not judge, so far did not mastered it.

Anton-Latukha:
Ok.

I feel uneasy merging everything. But someone needs to.

Issue #116, opened 2021-01-14
Dependency warnings


Anton-Latukha:
For Core I recieved:
```
[6 of 6] Compiling Main             ( tests/Driver.hs, /home/pyro/src/haskell/hnix-store/hnix-store-core/dist-newstyle/build/x86_64-linux/ghc-8.10.3/hnix-store-core-0.4.0.0/t/format-tests/build/format-tests/format-tests-tmp/Main.o ) [Test.Tasty.HUnit changed]

<no location info>: warning: [-Wunused-packages]
    The following packages were specified via -package or -package-id flags,
    but were not needed for compilation:
      - tasty-discover-4.2.2
Linking /home/pyro/src/haskell/hnix-store/hnix-store-core/dist-newstyle/build/x86_64-linux/ghc-8.10.3/hnix-store-core-0.4.0.0/t/format-tests/build/format-tests/format-tests ...

<no location info>: warning: [-Wunused-packages]
    The following packages were specified via -package or -package-id flags,
    but were not needed for compilation:
      - tasty-discover-4.2.2
Running 1 test suites...
Test suite format-tests: RUNNING...
Test suite format-tests: PASS
```

For Remote:
```
5 of 5] Compiling Main             ( tests/Driver.hs, /home/pyro/src/haskell/hnix-store/hnix-store-remote/dist-newstyle/build/x86_64-linux/ghc-8.10.3/hnix-store-remote-0.4.0.0/t/hnix-store-remote-tests/build/hnix-store-remote-tests/hnix-store-remote-tests-tmp/Main.o )

<no location info>: warning: [-Wunused-packages]
    The following packages were specified via -package or -package-id flags,
    but were not needed for compilation:
      - tasty-hunit-0.10.0.3
      - tasty-discover-4.2.2
      - base64-bytestring-1.2.0.1
Linking /home/pyro/src/haskell/hnix-store/hnix-store-remote/dist-newstyle/build/x86_64-linux/ghc-8.10.3/hnix-store-remote-0.4.0.0/t/hnix-store-remote-tests/build/hnix-store-remote-tests/hnix-store-remote-tests ...

<no location info>: warning: [-Wunused-packages]
    The following packages were specified via -package or -package-id flags,
    but were not needed for compilation:
      - time-1.9.3
      - tasty-hunit-0.10.0.3
      - tasty-discover-4.2.2
      - bytestring-0.10.12.0
      - binary-0.8.8.0
      - base64-bytestring-1.2.0.1
[...]

PR #117, opened 2021-01-14
Core, Remote: clean-up deps

Anton-Latukha:
Closing: #116
#115 allowed to reduce some deps in descriptions.


Anton-Latukha:
Also, `bytestring` mentions were duplicated.

Issue #118, opened 2021-01-14
Close the tasty-discover warnings


Anton-Latukha:
Core and Remote during compilation produce:
```
<no location info>: warning: [-Wunused-packages]
    The following packages were specified via -package or -package-id flags,
    but were not needed for compilation:
      - tasty-discover-4.2.2
```

The Internet search for me has not returned anything.
Which may indicate that the case is specific to configuration/use.
[Current documentation](https://hackage.haskell.org/package/tasty-discover)

Anton-Latukha:
`tasty-discover` seems to be needed as a system executable.

Because it is not included in Nix closure as such - the workaround of that was to put it in the `.cabal` file, so it now produces the warning, since it should not be in Cabal description.

Anton-Latukha:
BTW docs still does not instruct that `tasty-discover` is a run-time executable dependency.
And seems like report I created people bump into https://github.com/haskell-works/tasty-discover/issues/4.

Seems like not only I encountered this.

PR #119, opened 2021-01-14
Clean-up deps

Anton-Latukha:
* `cryptohash-sha512` was treated in Hackage by trustees, removing override.
* Closes: #118
`tasty-discover` is not needed in `.cabal`. If one removes the installed in the env `tasty-discover` executable would receive the situation I was puzzled about when I was building Store projects: https://github.com/haskell-works/tasty-discover/issues/4, since `tasty-discover` is if it is modules are not used - it is the runtime executable dependency which needs to be supplied externally from env.
  When I created that report - the `tasty-discover` was in `.cabal`, and I mention that in that text.
  
  So now it is a runtime executable dependency, and there seems to be no support for executable dependencies in Cabal. There is only one way to support it - install it in the environment.
  And Nix can support that through derivation expansion, which is done here.

### DIFF
--- a/hnix-store-core/hnix-store-core.cabal
+++ b/hnix-store-core/hnix-store-core.cabal
@@ -99,12 +99,10 @@ test-suite format-tests
     , binary
     , bytestring
     , containers
-    , filepath
     , directory
     , filepath
     , process
     , tasty
-    , tasty-discover
     , tasty-golden
     , tasty-hspec
     , tasty-hunit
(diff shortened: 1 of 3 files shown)


Anton-Latukha:
`tasty-discover` in Nixpkgs seems to be declared with `addBuildTool`, which seems to add it into Nixpkgs build closure. That should explain why there is no overrides to add `tasty-discover` in configurations there.
So this merge seems like does not need changes in Nixpkgs.

Issue #124, opened 2021-01-16
Please, close the Internal modules


Anton-Latukha:
Core does:
```
exposed-modules:
                     , System.Nix.Internal.Base32
                     , System.Nix.Internal.Hash
                     , System.Nix.Internal.Nar.Parser
                     , System.Nix.Internal.Nar.Streamer
                     , System.Nix.Internal.Nar.Effects
                     , System.Nix.Internal.Signature
                     , System.Nix.Internal.StorePath
```

They are Internal for purpose, and also `System.Nix.Internal` are nor used in the Remote, nor in the HNix.
Can we close them? I guess - yes we do.

PR #125, opened 2021-01-16
Core: Internal modules are internal

Anton-Latukha:
Closes: https://github.com/haskell-nix/hnix-store/issues/124

This internal API is not used in Remote or HNix.


Anton-Latukha:
Ok, learned something new.

sorki:
I wanted to do something similar for `-remote` but realized it's not that easy since tests can't import `other-modules`.

PR #126, opened 2021-01-16
Forming release: updating changelogs, updating cabal versions

Anton-Latukha:
Went and checked the changelog to make it into release.

Looked through diffs: https://github.com/haskell-nix/hnix-store/compare/0.4.0.0...master
Dependency enumeration is superfluous info to the user - just stated "Big deps clean-up", aka "... it is good!" thing.
The only change that is not mentioned is: [pointer](https://github.com/haskell-nix/hnix-store/compare/0.4.0.0...master#diff-9aba4fa73b66a8cc405b0bb3b8e64d92d918ae79377ff0556e60bdd78b4464b5), that is an internal change, hardly people are interested to hear it.

All other changes are changes in the test suites or purely cosmetic refactors.

Seems like `0.4.0.1` would be too little - too many changes overall, and remote got API expantion, `0.4.1.0` is safe to bless both cases for simplicity.

### DIFF
--- a/hnix-store-core/ChangeLog.md
+++ b/hnix-store-core/ChangeLog.md
@@ -1,8 +1,8 @@
 # Revision history for hnix-store-core
 
-## [next](https://github.com/haskell-nix/hnix-store/compare/0.4.0.0...master) 2021-MM-DD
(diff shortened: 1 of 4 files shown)


sorki:
Thank you, much appreciated, I'll release it right away.

PR #132, opened 2021-01-31
Help HLS to work-out monorepo: add HLS Cradle config, rm Setup.hs from Core

Anton-Latukha:
The HLS had a problem that it was detecting the `cabal.project` in root of our
monorepo, but also was founding the Setup.hs that promised that it is a simple
project structure. And was trying to run the Core with `runhaskell`.

So HLS was throwing several errors. The explicit definition of cradles allowed
HLS to understand that it is a monorepo. Removing of Setup.hs is
because HLS sometimes was trying to run the Core with `runhaskell`.
and Setup.hs is not recommended anymore in the docs, and upstream work plans to deprecate it
completely in favor of a proper solution.

### DIFF
--- a/hie.yaml
+++ b/hie.yaml
@@ -1,0 +1,13 @@
+cradle:
+  cabal:
+    - path: "./hnix-store-core/src"
+      component: "lib:hnix-store-core"
+
+    - path: "./hnix-store-core/tests"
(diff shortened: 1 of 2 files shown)


Anton-Latukha:
Yep.

HLS `0.9` sometimes does not start when `Setup.hs` is present

PR #134, opened 2021-02-02
Core, Remote: handcrafted clean-up

Anton-Latukha:
First, `brittany` was used.

All changes that `brittany` were suggesting were passed through manual supervision.

Then handcrafted code cleanup was done: basic: indentation & sectioning, removal of the `do` where they are vacuous, `bool`, `either` use.

All changes are pure lambda code refactoring, there should be no changes to the functionality.

### DIFF
--- a/hnix-store-core/Setup.hs
+++ b/hnix-store-core/Setup.hs
@@ -1,2 +1,2 @@
-import Distribution.Simple
+import           Distribution.Simple
 main = defaultMain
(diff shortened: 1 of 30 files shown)


Anton-Latukha:
Ok.

I did pretty clean and checked work here, and checked everything several times over during work and after publishing.

I have time to do more.

I probably need moving and ask and wait for reviews on real-deal code logic changes.

Issue #121, opened 2021-01-15
Please, suit the testsuites to the environement they support


Anton-Latukha:
We pretty frequently lately have a need to: overload `.cabal` with flags, disable some tests on macOS, add/remove Nix overrides for tests, and go into Nixpkgs to maintain tests, to fix some specific test that fails and not suited for some environment.

We can just check in the code the environment where tests are running and simply not run the tests that we know are not suited for that platform. That would allow us to stop flounce around configs and custom commands goalkeeping the tests.

This is projects-wide idea.

The check for macOS is simple.
`nix-build` seems to obscure from programs where they build in, but I thing something like:

```haskell
  execPath <- fmap (fromMaybe "") $ findExecutable "bash"
  if (isInfixOf "nix" $ execPath) && (isInfixOf "store" $ execPath)
    then (putStrLn $ show execPath)
    else error "Not in nix-build"
```
--- that tells that we run in `nix-build`, `NixOS` already would report other path, so we detected the `nix-build`.

After that is automated - all we would need to do - is to do is to check the PR in CI and merge the PR in green CI, and rightfully inform - mention the test specifics in the Readme, like "To run `io-testsuite` in Remote - install/run Nix in daemon multiuser mode", and if people would not read it - no biggy.
And we would run CI with reference development requirements (which we nee there anyway) and check PRs through CI, so if in development someone had no ability to run tests for something - CI tests it for people.

Anton-Latukha:
Got the environment from the `nix-build`, in fact there is a lot Nix-specific variables in the environment:
```
NIX_BINTOOLS=/nix/store/2xrj4wg0sn53lbvi6qgkv3vyyzf8c39a-binutils-wrapper-2.34
NIX_BINTOOLS_FOR_BUILD=/nix/store/2xrj4wg0sn53lbvi6qgkv3vyyzf8c39a-binutils-wrapper-2.34
NIX_BINTOOLS_WRAPPER_TARGET_BUILD_x86_64_unknown_linux_gnu=1
NIX_BINTOOLS_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu=1
NIX_BUILD_CORES=8
NIX_BUILD_TOP=/build
NIX_CC=/nix/store/y78inja64xaaq6bx33nxn2h1lgw0jswk-gcc-wrapper-10.2.0
[...]

PR #136, opened 2021-02-24
core: fix nar test failing when there's no /proc

sorki:
Fixing the most pressing one.

### DIFF
--- a/hnix-store-core/tests/NarFormat.hs
+++ b/hnix-store-core/tests/NarFormat.hs
@@ -231,7 +231,7 @@ test_streamManyFilesToNar = HU.testCaseSteps "streamManyFilesToNar" $ \step ->
         IO.withFile "hnar" IO.WriteMode $ \h ->
           buildNarIO narEffectsIO narFilePath h
         filesPostcount <- countProcessFiles
-        pure $ filesPostcount - filesPrecount
+        pure $ (-) <$> filesPostcount <*> filesPrecount
 
     step "create test files"
     Directory.createDirectory packagePath
@@ -252,7 +252,9 @@ test_streamManyFilesToNar = HU.testCaseSteps "streamManyFilesToNar" $ \step ->
 
     step "check constant file usage"
     filesPostcount <- countProcessFiles
-    (filesPostcount - filesPrecount) `shouldSatisfy` (< 50)
+    case ((-) <$> filesPostcount <*> filesPrecount) of
+      Nothing -> pure ()
+      Just c -> c `shouldSatisfy` (< 50)
 
     -- step "check file exists"
     -- e <- doesPathExist packagePath'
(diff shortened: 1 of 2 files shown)


hnix-store-core/tests/NarFormat.hs
```diff
@@ -252,7 +252,9 @@ test_streamManyFilesToNar = HU.testCaseSteps "streamManyFilesToNar" $ \step ->
 
     step "check constant file usage"
     filesPostcount <- countProcessFiles
-    (filesPostcount - filesPrecount) `shouldSatisfy` (< 50)
+    case ((-) <$> filesPostcount <*> filesPrecount) of
+      Nothing -> pure ()
+      Just c -> c `shouldSatisfy` (< 50)
```
Anton-Latukha, line 257: It is very often encountered.

I tend to use `maybe` for it.
Results in a less hardcoded structure.
Idea is that pattern matching is a hardcode based on the type with a space-structuring coupling.
If no pattern matching, not space structure dependency - no hardcode and layout can be done in any useful way.

Because of that `bool` can be easily promoted into `maybe, & `maybe` to `either` & also their code layout is seamless.

What found - is that patterns in the code structure easier to see. Factoring-out the same structure is also easier.

So far it worked quite well.

hnix-store-core/tests/NarFormat.hs
```diff
@@ -361,12 +363,16 @@ packThenExtract testName setup =
...
-  let fdDir = "/proc/" <> show pid <> "/fd"
-  fds  <- P.readProcess "ls" [fdDir] ""
-  pure $ length $ words fds
+  hasProc <- doesDirectoryExist "/proc"
+  if not hasProc
+    then pure Nothing
+    else do
+      let fdDir = "/proc/" <> show pid <> "/fd"
+      fds  <- P.readProcess "ls" [fdDir] ""
+      pure $ pure $ length $ words fds
```
Anton-Latukha, line 375: `pure` can be factored-out, `Nothing` is `mempty`. It would be seen that what return in previous structure with the internal value computed here.

It may not look as important at the moment, but over time.

Anton-Latukha:
These are just simple comments.
Those comments are not important.

You did a better job than my idea, wanted to do mac-specific detection, but if `/proc` is used - test that it exists - is much proper programmers way.

I need to learn to look to at platform-specific reports from the code standpoint.

Anton-Latukha:
Would merge to keep the ball rolling.

Be free to self-merge simple fixes like this.

We are not a slow huge enterprise. Out strength is the ease to do refactoring, development & overall flexibility & speed of the position.

For example, if you'd self-merged yesterday the fix - it would've been published to the Hackage in the same sitting and so - people would receive it and you can close the reports.

We roughly know when the change is such that the cross-review is needed.

Those are some words, they can be ignored and would stay in the stage of air shakers.

sorki:
I do find these helpful as sometimes I miss some obvious ones as `mempty` / `maybe`.

I was wondering if I should do something like test prerequisites as outlined in #121 but in this case it looked better to switch the type of the function to `Maybe` to account for failure.

Anton-Latukha:
Well,  #121 was opened as a discussion ground to create a separate project for that.

Opened the topic, because there is none currently, and especially at current times and known future things the package probably would be quite popular. Without it programmers need to pass the system information manually to detect the platform program runs on.

But as I see, it is better to make the code work with the environment - it is a more straight programmer thinking.

PR #137, opened 2021-02-24
Release scripts

sorki:
Wrote this recently to be able to release both at once, related to #101. Needs work.


sorki:
Treated as separate packages. Closing.

PR #144, opened 2021-03-23
add a readonly computeStorePathForPath

layus:


layus:
It seems that it does not work correclty, for now. I do not get the right hash in the end :-/

layus:
Fixed in https://github.com/haskell-nix/hnix-store/pull/145 :tada:

layus:
I do not understand why I need AllowAmbiguousTypes and TypeFamilies here. Looks way overkill for that simple function.

layus:
Just to give an idea of what I am facing now:

```
src/System/Nix/ReadonlyStore.hs:76:28: error:
    • Could not deduce (NamedAlgo a0)
      from the context: NamedAlgo a
        bound by the type signature for:
                   computeStorePathForPath :: forall (a :: HashAlgorithm).
                                              NamedAlgo a =>
                                              StorePathName
                                              -> FilePath
                                              -> Bool
                                              -> (FilePath -> Bool)
                                              -> Bool
                                              -> IO StorePath
        at src/System/Nix/ReadonlyStore.hs:(76,28)-(82,26)
      The type variable ‘a0’ is ambiguous
    • In the ambiguity check for ‘computeStorePathForPath’
      To defer the ambiguity check to use sites, enable AllowAmbiguousTypes
      In the type signature:
        computeStorePathForPath :: forall a.
                                   (NamedAlgo a) =>
                                   StorePathName
                                   -> FilePath -> Bool -> (FilePath -> Bool) -> Bool -> IO StorePath
   |
76 | computeStorePathForPath :: forall a. (NamedAlgo a)
   |                            ^^^^^^^^^^^^^^^^^^^^^^^...
```

layus:
Fixed everything :-). I think it is ready
@sorki Would you consider reviewing it ?

hnix-store-core/src/System/Nix/ReadonlyStore.hs
```diff
@@ -45,9 +47,7 @@ makeTextPath fp nm h refs = makeStorePath fp ty h nm
 
 makeFixedOutputPath
   :: forall hashAlgo
-  . ( ValidAlgo hashAlgo
-    , NamedAlgo hashAlgo
-    )
+  . (NamedAlgo hashAlgo)
```
sorki, line 50: Fixed superfluous parens here and pushed to master.

sorki:
Thank you!

layus:
I copied the style from makeStorePath above :-)

Le 24 mars 2021 09:22:12 GMT+01:00, Richard Marko ***@***.***> a écrit :
***@***.*** commented on this pull request.

hnix-store-core/src/System/Nix/ReadonlyStore.hs
```diff
@@ -69,3 +69,22 @@ makeFixedOutputPath fp recursive h =
 computeStorePathForText
   :: FilePath -> StorePathName -> ByteString -> (StorePathSet -> StorePath)
 computeStorePathForText fp nm = makeTextPath fp nm . hash
+
+computeStorePathForPath
+  :: StorePathName        -- ^ Name part of the newly created `StorePath`
+  -> FilePath             -- ^ Local `FilePath` to add
+  -> Bool                 -- ^ Add target directory recursively
+  -> (FilePath -> Bool)   -- ^ Path filter function
+  -> Bool                 -- ^ Only used by local store backend
```
sorki, line 78: I feel like we should do something about these two arguments but I'm not sure what.

sorki:
Oops, I'm opening style can of worms discussion. Wanted to do that anyway, maybe we can ormolu everything and forget about such.

layus:
> maybe we can <anything> everything and forget about such.

Yes, I spend a lot of time indenting, formatting and similarly torturing myself to get formatting right, even though I do not have any strong opinion on the right formatting. Any automated tool would get rid of this wasted time.

PR #145, opened 2021-03-24
Reorder references in makeTextPath

layus:
It looks like this is a longstanding bug, due to changes in the set representation over the time.


sorki:
Good catch!

Issue #156, opened 2021-06-07
Please, refactor the Hash: subtasks


Anton-Latukha:
Continuation of: #92

1. Separate the `Base` encoding code from the hashing code.
2. Abstract NixBase32. `Base32` package || a separate package with `Base{16,32,64}` compatible interface: https://github.com/emilypi/Base32/issues/12
2. Use the `Base{16,32,64}` packages. Break everything that does not adhere to the change.
2. Separate `sri-hash` into a separate package, with a hash-types compatible interface, write documentation on it. Use the package & break everything that does not adhere to the change.
2. `Truncated 20 SHA256` is the only truncated type we need.
2. Break all other things, until it works & uses `cryptonite <|> hashing`...
2. Live eternally.
2. Profit!

Plan can not be circumvented.

Anton-Latukha:
Wow, this is *clooosed*, thank you everybody for attention, see - the curtain is falling.

Anton-Latukha:
Well, did I implemented task №7?

PR #157, opened 2021-06-07
Refactors 2021-06-07

Anton-Latukha:
Small tweaks & refactors & reorganizations & so - breakages towards flexiblity & easier: #156

~Working on Core compatibility with Remote breaks but `¯\_(ツ)_/¯` what would one do. When I work on particular things I do not want to be forced to do sweeping changes across the stack, prefer to release Core & after that migrate the Remote & release it right after if needed, somewhere in time there is a time to stop synchronously tagging them in the same commit.~
Hey, CI & Nix config - it all works now https://github.com/haskell-nix/hnix-store/pull/160

### DIFF
--- a/hnix-store-core/ChangeLog.md
+++ b/hnix-store-core/ChangeLog.md
@@ -1,9 +1,41 @@
-# Revision history for hnix-store-core
+# ChangeLog
+
+## WIP
+
+* Breaking:
+
+  * [(link)](https://github.com/haskell-nix/hnix-store/pull/157/commits/97146b41cc87327625e02b81971aeb2fd7d66a3f) Migration from packages `cryptohash-` -> `cryptonite`:
+    `System.Nix.Hash`:
+      * rm `data HashAlgorithm` in favour of `cryptonite: class HashAlgorithm`
+      * rm `class ValidAlgo` in favour of `cryptonite: class HashAlgorithm`.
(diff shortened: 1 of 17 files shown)


Anton-Latukha:
https://github.com/haskell-nix/hnix-store/pull/161#issuecomment-858529060
> Looking good. I wonder - was `ValidAlgo` superfluous?

A lot kinda. I'd state it was quite an "in advance" type system overfitting.

As I understand, `ValidAlgo` was created to have a list of hash algorithms & to gatekeep the list of "valid hashing algorithms that are supported".

But all the code is polymorphic on the hashing algorithms, which also have 1 interface. Which imposed a question: why is there is need to gatekeep hash algorithms, we actually use only `SHA256`, but also just as easy can allow people to us ANY hashing algorithms if they want.
Removing `ValidAlgo` both simplified the type system & also allowed the code to work for all of the hashing algorithms provided by `cryptonite` library.

But, `class NamedAlgo` only has instances that were enumerated in the old `HashAlgorithm`. In fact `NamedAlgo` exists only because hashing libraries abstain from providing the canonical names for the algorithms, well they are also right, because industry made it impossible to give  canonical names: is it SHA256 or SHA-256 or SHA2 or SHA-2?

So now we both allow to use any hashing alg, with a small caveat, if they every would use `class NamedAlgo`, they would need to provide an instance with their way of naming for the algorithm in use.

PR #161, opened 2021-06-10
Remote: support of Core migration to cryptonite

Anton-Latukha:


sorki:
Looking good. I wonder - was `ValidAlgo` superfluous?

layus:
> Looking good. I wonder - was `ValidAlgo` superfluous?

Well The only algo that was valid but not named was `Truncated n`. So if we remove the Truncated hash type, there is little point in keeping both. As we also have the implicit assumption that every named algo is valid, There is no point in keeping both.

But the class NamedAlgo makes less sense now. Maybe rename NamedAlgo -> HashAlgo ?

Anton-Latukha:
Sorry if commit least mislead, I created a brunch out of #157 with Remote update on top of it.

It was done to fix the CI for #157, by a way of #160.

In other words - this PR is 1 commit, and brunch (now after #160) would work & CI would build only after the Core release.

hnix-store-remote/src/System/Nix/Store/Remote/Builders.hs
```diff
@@ -38,4 +35,4 @@ contentAddressableAddressBuilder (Fixed _narHashMode (SomeDigest (digest :: Dige
 
 digestBuilder :: Digest a -> Builder
 digestBuilder =
-  Data.Text.Lazy.Builder.fromText . System.Nix.Hash.encodeInBase Base32
+  Data.Text.Lazy.Builder.fromText . System.Nix.Hash.encodeDigestWith NixBase32
```
Anton-Latukha, line 38: Need to fix this

Ericson2314:
I forget the details, sadly, but I've heard of a number of people having a bad experience with cryptonite.

Anton-Latukha:
> I forget the details, sadly, but I've heard of a number of people having a bad experience with cryptonite.

Well, it is a WAY better experience than with `cryptohash` or `cryptohash-*`. 

Since in `cryptonite/memory/basement/foundation` Vincent, with a delay, but at least shown-up to support GHC 9.0.

"Bad experience with cryptonite" depends to "where in the library" & on the angle & the degree. & We currently not running BigData silo hashmaps here. How I know it, Nix hashing is hashing of a fingerprint of small textual data, so I guess (it can be said) hashing does not have an impact on the overall performance.

These migrations also mainly impact the HNix itself, so these transitions are mostly abstracted away from users.

The migration was done to the `cryptonite` as to the current standard for hashing. Before making the obvious switch, I actually again checked at what hashing libraries are there & the answer is: "not many" & mainly [`hashing`](https://hackage.haskell.org/package/hashing), which is still a sloth.

Anton-Latukha:
Well, here we meet another constraint of the CI - the Core 0.5 release happened, but it is not yet imported into Nixpkgs.

PR #167, opened 2021-10-30
Add `GHC.Stats` import to NarFormat test

hololeap:
This is needed for the tests to compile on my system.
Without the import, `max_live_bytes` and `getRTSStats` are undefined.


Anton-Latukha:
Thanks for sending the patch.

The expression "for the tests to compile on my system" seems uninformative. Please, can you add in the code a little bit of explanation/exemplation, why it may be needed for some cases? Maybe CPU architecture & OS at least, maybe build is configured with `rtsopts`? From that, a lot can follow.

PR #172, opened 2021-11-15
remote/tests/NixDaemon: use Text

Anton-Latukha:


sorki:
Rebased for CI.

sorki:
Note to self: this needs `relude 1.0` override in nixpkgs after release similar to hnix.

PR #177, opened 2022-02-08
replace the filepath argument in `addToStore` with a more common type `NarSource` (The core part)

### LINKED ISSUE #176: Current `addToStore` API from hnix-store is too limited, we should shape it
To implement `fetchurl`  in hnix, haskell-nix/hnix#1050, haskell-nix/hnix#1051.

We need to perform the following.
Dowload the file content -> dump file content to nar -> Add the nar to remote store.

But in current `addToStore` api of hnix-store we can just send local file to daemon

soulomoon:
`type NarSource m =  (ByteString -> m ()) -> m ()`.
* `FilePath` can turn to `NarSource m`.
* `Text` can turn to `NarSource m`

So It could fix #176

### DIFF
--- a/hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
+++ b/hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
@@ -66,7 +66,7 @@ streamNarIO effs basePath yield = do
       fSize <- IO.liftIO $ Nar.narFileSize effs path
       yield $ str "contents"
       yield $ int fSize
-      yieldFile path fSize 
+      yieldFile path fSize
 
     when isDir $ do
       fs <- IO.liftIO (Nar.narListDir effs path)
--- a/hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
+++ b/hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
@@ -3,7 +3,9 @@
 {-# language ScopedTypeVariables #-}
 
 module System.Nix.Internal.Nar.Streamer
-  ( streamNarIO
+  ( NarSource 
+  , dumpString
+  , streamNarIO
   , IsExecutable(..)
   )
 where
@@ -19,22 +21,49 @@ import           System.FilePath                  ( (</>) )
 import qualified System.Nix.Internal.Nar.Effects as Nar
 
 
+int :: Integral a => a -> ByteString
+int n = Serial.runPut $ Serial.putInt64le $ fromIntegral n
+
+str :: ByteString -> ByteString
+str t =
+  let
+    len = Bytes.length t
+  in
+    int len <> padBS len t
+
+padBS :: Int -> ByteString -> ByteString
+padBS strSize bs = bs <> Bytes.replicate (padLen strSize) 0
+
+strs :: [ByteString] -> ByteString
+strs xs = Bytes.concat $ str <$> xs
+
+-- | NarSource 
+-- the source to provide nar to the handler
+type NarSource m =  (ByteString -> m ()) -> m ()
+
+-- | dumpString
+-- dump a string to nar
+dumpString :: forall m . IO.MonadIO m => ByteString -> NarSource m
+dumpString text yield = do
+  yield $ str "nix-archive-1"
+  yield $ str "("
+  yield $ str "type" 
+  yield $ str "regular"
+  yield $ str "contents"
+  yield $ str text
+  yield $ str ")"
+
+
 -- | This implementation of Nar encoding takes an arbitrary @yield@
 --   function from any streaming library, and repeatedly calls
 --   it while traversing the filesystem object to Nar encode
-streamNarIO
-  :: forall m
-   . (IO.MonadIO m)
-  => (ByteString -> m ())
-  -> Nar.NarEffects IO
-  -> FilePath
-  -> m ()
-streamNarIO yield effs basePath = do
+streamNarIO :: forall m . IO.MonadIO m => Nar.NarEffects IO -> FilePath -> NarSource  m
+streamNarIO effs basePath yield = do
   yield $ str "nix-archive-1"
-  parens $ go basePath
-
+  yield $ str "("
+  go basePath
+  yield $ str ")"
  where
-
   go :: FilePath -> m ()
   go path = do
     isDir     <- IO.liftIO $ Nar.narIsDir effs path
@@ -53,7 +82,7 @@ streamNarIO yield effs basePath = do
       fSize <- IO.liftIO $ Nar.narFileSize effs path
       yield $ str "contents"
       yield $ int fSize
-      yieldFile path fSize
+      yieldFile path fSize 
 
     when isDir $ do
       fs <- IO.liftIO (Nar.narListDir effs path)
@@ -64,16 +93,6 @@ streamNarIO yield effs basePath = do
           let fullName = path </> f
           yield $ strs ["name", Bytes.Char8.pack f, "node"]
           parens $ go fullName
-
-  str :: ByteString -> ByteString
-  str t =
-    let
-      len = Bytes.length t
-    in
-      int len <> padBS len t
-
-  padBS :: Int -> ByteString -> ByteString
-  padBS strSize bs = bs <> Bytes.replicate (padLen strSize) 0
 
   parens act = do
     yield $ str "("
@@ -86,13 +105,7 @@ streamNarIO yield effs basePath = do
   yieldFile path fsize = do
     mapM_ yield . Bytes.Lazy.toChunks =<< IO.liftIO (Bytes.Lazy.readFile path)
     yield $ Bytes.replicate (padLen $ fromIntegral fsize) 0
-
-  strs :: [ByteString] -> ByteString
-  strs xs = Bytes.concat $ str <$> xs
-
-  int :: Integral a => a -> ByteString
-  int n = Serial.runPut $ Serial.putInt64le $ fromIntegral n
-
+          
 
 data IsExecutable = NonExecutable | Executable
   deriving (Eq, Show)
--- a/hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
+++ b/hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
@@ -5,6 +5,7 @@
 module System.Nix.Internal.Nar.Streamer
   ( NarSource 
   , dumpString
+  , dumpPath
   , streamNarIO
   , IsExecutable(..)
   )
@@ -25,6 +26,7 @@ import qualified System.Nix.Internal.Nar.Effects as Nar
 -- the source to provide nar to the handler
 type NarSource m =  (ByteString -> m ()) -> m ()
 
+
 -- | dumpString
 -- dump a string to nar
 dumpString :: forall m . IO.MonadIO m => ByteString -> NarSource m
@@ -36,6 +38,13 @@ dumpString text yield = do
   yield $ str "contents"
   yield $ str text
   yield $ str ")"
+
+
+-- | dumpPath
+-- shorthand
+-- build a Source that turn file path to nar using the default narEffectsIO.
+dumpPath :: forall m . IO.MonadIO m => FilePath -> NarSource m
+dumpPath = streamNarIO Nar.narEffectsIO
 
 
 -- | This implementation of Nar encoding takes an arbitrary @yield@
--- a/hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
+++ b/hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
@@ -21,22 +21,6 @@ import           System.FilePath                  ( (</>) )
 import qualified System.Nix.Internal.Nar.Effects as Nar
 
 
-int :: Integral a => a -> ByteString
-int n = Serial.runPut $ Serial.putInt64le $ fromIntegral n
-
-str :: ByteString -> ByteString
-str t =
-  let
-    len = Bytes.length t
-  in
-    int len <> padBS len t
-
-padBS :: Int -> ByteString -> ByteString
-padBS strSize bs = bs <> Bytes.replicate (padLen strSize) 0
-
-strs :: [ByteString] -> ByteString
-strs xs = Bytes.concat $ str <$> xs
-
 -- | NarSource 
 -- the source to provide nar to the handler
 type NarSource m =  (ByteString -> m ()) -> m ()
@@ -57,7 +41,7 @@ dumpString text yield = do
 -- | This implementation of Nar encoding takes an arbitrary @yield@
 --   function from any streaming library, and repeatedly calls
 --   it while traversing the filesystem object to Nar encode
-streamNarIO :: forall m . IO.MonadIO m => Nar.NarEffects IO -> FilePath -> NarSource  m
+streamNarIO :: forall m . IO.MonadIO m => Nar.NarEffects IO -> FilePath -> NarSource m
 streamNarIO effs basePath yield = do
   yield $ str "nix-archive-1"
   yield $ str "("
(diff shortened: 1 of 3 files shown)


soulomoon:
The conversion name is still need to be revised

soulomoon:
path filter should take place in building narsource, remove from addToStore

soulomoon:
The workflows failed. 
It seems the dependency of hnix-store-remote depend on release version of hnix-store-core, but the changes spans over both repository.

sorki:
Yes, that's intentional, see #101 - they are treated as separate packages and we need to first tackle `core` part PR, make a release and do the same for remote.

soulomoon:
cool, understand.

Anton-Latukha:
*ifirc* the `core` need to be updated & released before merging changes into remote. I waited, maybe you would do 2 PRs out of this one,

Or you consider this PR ready. The PR never was marked a draft, so seems its status never transmitted consideration of the readiness.

Is this PR considered ready?

soulomoon:
@Anton-Latukha , The updated remote and updated core is compatible as I tested in local. But the current PR would break the pipeline. I would stash the remote changes and make the pipelines work. Then after this PR is merged, pop the stash to make another PR.

Anton-Latukha:
Ok. Today is late, would attend tomorrow, if the situation would allow.

I'd preferred @sorki to review `hnix-store` things, because I know part of `hnix-store`, but `sorki` has deeper knowledge of it & of Nix processes there.

And also the more I involve with `hnix-store` - the more I would take responsibility for `hnix-store` the more I would want to refactor it to the layout form that HNix is currently in. I'd loved to move `Nix.Utils` under all Nix projects & use some of the most helpful ones in the `store` code. I try to abstain from doing refactoring and with `sorki` we kind of talked about it back in the day.

hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
```diff
@@ -19,22 +22,41 @@ import           System.FilePath                  ( (</>) )
...
 
+-- | NarSource 
+-- the source to provide nar to the handler
+type NarSource m =  (ByteString -> m ()) -> m ()
+
+
+-- | dumpString
+-- dump a string to nar
+dumpString :: forall m . IO.MonadIO m => ByteString -> NarSource m
+dumpString text yield = do
```
Anton-Latukha, line 33: This is strange. Even confusing to read.

Type signature has 1 argument, but the function has 2.

I'd argumented that it is more confusing to read than my style. Because if more of such types are used - that would make type signatures unreadable & so not useful for programmers.

hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
```diff
@@ -19,22 +22,41 @@ import           System.FilePath                  ( (</>) )
...
+  yield $ str "nix-archive-1"
+  yield $ str "("
+  yield $ str "type" 
+  yield $ str "regular"
+  yield $ str "contents"
+  yield $ str text
+  yield $ str ")"
```
Anton-Latukha, line 40: Also, seems that `yield` is a stable function that would get reused a lot in the function and between function runs, while `text` is a more dynamic information, for different function runs `text` frequently would be always different.

So it has a sense to put arguments:
```haskell
dumpString :: forall m . IO.MonadIO m => NarSource m -> ByteString -> ..
```

It would ensure that GHC infers to compile-in & cache `yeald` into memory stack.
(this argument is a contrived one, because for functions with 2 arguments GHC always guarantee to infer most optimal argument order for the uses).

But that also puts the signature & code into a canonical Haskell order.

This `yeald` repeat block is: `traverse_ (yeald . str) [ "nix-archive-1", ...]`.

Anton-Latukha: Then the documentation would mention that it takes a getter/MonadIO printer for text & monad prints the Nar form.

You also can use type signature argument documentation for that, that is a good case to explain this type of signature.

ArgDocs are not yet shown, but I am actively working on it in HLS, to provision the Haddock argument documentation to programmers.

With reading along with the type signature with attached notes for every entity - person reads line by line how to work with the function & understands its internal semantic enough at the same time so does not becomes puzzled & so it saves the need for programmers look into the function implementation source, which is the ideal main agenda of documentation - to have almost no documentation & yet it explaining the complex topics & forming an intuitive understanding of the code without the need to go into reading the code that is in the other place.

hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
soulomoon: Agree that it is confusing to read, especially with remote part getting stashed.
Let me justify the this style.

It is done in CPS style. Under Yoneda Lemma in category theory, `NarSource m` is isomorphic to `ByteString`.
It is to align with code already in the codebase.
```haskell
runOpArgsIO
  :: WorkerOp
  -> ((Data.ByteString.ByteString -> MonadStore ()) -> MonadStore ())
  -> MonadStore ()`. 
```
Using this style instead of directly `ByteString`, I think is it to enable to chunks the `IO` input instead of feeding the whole nar at once to achieve better performance. 

It would be more reasonable once we look at the API of addToStore. To generify the the original API to more than taking in a `FilePath`, we need the `NarSource` to take a handler `ByteString -> m ()` to actually process the nar bytes.

hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
soulomoon: Thank you for the suggestions.

* `dumpString` won't be getting the yield function until the last moment CPS style `runOpArgsIO` in the existing codebase to feed it with the yield. Thus dumpString however would need to accept the text first.
* I like this `traverse_ (yeald . str) [ "nix-archive-1", "(", .. ")" ]`. I should make the change.
* I lack experience of using the argdocs and do not know how to do it. May be you could show me.

hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
soulomoon: The Remote Part is something like this:

```haskell
-- | Pack `Nar` and add it to the store.
addToStore1
  :: StorePathName        -- ^ Name part of the newly created `StorePath`
  -> NarSource MonadStore -- ^ provide nar stream
  -> Bool                 -- ^ Add target directory recursively
  -> (FilePath -> Bool)   -- ^ Path filter function
  -> RepairFlag           -- ^ Only used by local store backend
  -> MonadStore StorePath
addToStore1 name source recursive _pathFilter _repair = do
  runOpArgsIO AddToStore $ \yield -> do
    yield $ toStrict $ runPut $ do
      putText $ unStorePathName name
      putBool $ not $ algoName @Hash.SHA256 == "sha256" && recursive
      putBool recursive
      putText $ System.Nix.Hash.algoName @Hash.SHA256
    source yield
  sockGetPath
```

hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
Anton-Latukha: Oh, I was not aware of the CPS.

Honestly - I am yet to use it in practice myself. Also, I remember reading that Haskell has several alternatives to do the same, and that some of them in some cases even optimize much better, and also GHC is a "smart enough compile", but of course CPS is a very flexible well developed Lambda Calculus/Lisp derivatives practice I love CPS overall, but think in the majority of cases the tail arguments then need to be noted as CPS tail passing.

---

ArgDocs: It is simply using `-- ^` on function arguments. I call them ArgDocs - because it is how it is called in GHC API at least.

<sub>
Nota bene: By the way, which is interesting HLS mostly uses GHC to get documentation & plans to migrate fully. Recent versions of GHC provide enough functionality to sort-of replace `haddock` in a lot of cases. Comments are a part of the specification overall, for literate programming reasons it is also sane to have doc definitions in the language spec & so in ADT also. By the way, comments still are not implemented in HNix.
</sub>

</p>

HLS even allows to click on the function type signature declarations to do it, but it yet does not shows the result back.

---

Theme: Documenting arguments as one of the main sources of explanation.

Derives from the basic ideas.

The most frequent lazy style of writing documentation is: [`traverse_`](https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-Foldable.html#v:traverse_)
> traverse_ :: (Foldable t, Applicative f) => (a -> f b) -> t a -> f ()
> Map each element of a structure to an Applicative action, evaluate these actions from left to right, and ignore the results.

Lets do the same thing but generalize this to *reduction ad absurdum*:
> functionName :: (StrangeClass f, HeadSpinningClass g, Walkable h) f -> g -> h -> a -> b -> c
> Map each element of a structure to a StrangeClass f, evaluate these actions from left to right, or do the other given action if not - do the other one, and ignore the result.

People writing documentation may think people understand them, while documentation is completely puzzling. Maps over what. which actions in which order are meant in the doc.

The same we can observe even in `traverse_` if we squint and pretend we do not know something (people read the documentation when they do not know anything or something in type signature), so the doc says it maps what? each element (person not knowing - tries to apply the knowledge that people & docs tend to sequentially mention what gets takes & text entails that first - function accepts elements and maps them over the second argument), but the `traverse_` doc is written non-sequentially, so mapping of the doc to the type signature is left "for the user to practice understanding of how documentation maps to the API of the function". So person needs to go and learn all types present in type signature and internalize knowledge, of `(StrangeClass f, HeadSpinningClass g, Walkable h)` and then go back & try to understand how the doc description mention elements & what is implicit mapping of those elements to the API of the function.

Where in reality the `Map each element of a structure to an Applicative action, evaluate these actions from left to right, and ignore the results.`, the form of description: "function takes value to transform, takes transformation, takes takes accessor for the angle, returns tangent of an angle" what they do and meant to do - they mentally go over the type signature & describe what arguments are accepted & how they compose, that means they wanted to do:
```haskell
fun
  :: a -- ^ transformation
  -> b -- ^ getter for angle
  -> c  -- ^ value to transform
  -> d  -- ^ tangent
```

No handwaving what part of the text meant what argument - no inference of implicit map of internal knowledge, that requirement got eliminated simply laying the "takes this, takes that, does this with that" - to what people semantically described - an API & meaning of arguments it takes.

---

The `->` is a general intuitive symbol for action. Haskell is cool because of its functional approach (functions) & their type system. Type system implements a Hask "category". Functions are morphism that also denoted `->`, and `:: a -> b` in Haskell is action, computation, function, morphism.

During development, what defines the good function for the programmer from a random Haskell community place - a function that is easy to use.
To be usable - the function needs to be understandable to the point that person is sure what function does.

That means either function is without docs & considered understandable on its own (and we naively implicitly assume everyone is like us in understanding), or if there is documentation - it is better for it to effectively in a short amount of symbols convey what function does so it becomes clear to the reader.

The main flaws of documentation. 1. Semantic transmission, 2. Volume & effort-time to read it, 3. People not reading documentation because of all other mentioned entries. 4. During development - developers forget about documentation and so documentation rots and so user trust in documentation diminishes.

Now, from other perspective:

What function from the outside has.

The most theoretical case is:
```haskell
((\    a -> a) :: a -> a)
```

That is the minimal definition and API of a function. As we see - the most important part in the function is code, and then the type of a function.

The next essential is the name.

And that forms the API of the function: `name :: a -> b`. How to document the API? Precicely - document the `name` & `:: a -> b`. People tend to put everything into the main documentation block (explaining `name`) & forget the explain type signature which in fact describes the function, types as proofs, if person does not uses the type signature Haskell/GHC - would not allow to use the function/code. `name` can be explained as much as people want, but if in that reader does not understand type signature - the explanations & docs were useless, since person can not use the function in code.

It also can be seen that in `name :: a -> b`, the `a -> b` hold the symbol of a an action/function/morphism, `name` holds only a `name`, from `name` alone it is impossible to tell whenever it is a function or a value - person needs to look in type signature & see ` :: a -> b` - Ok - now I see is a function.

The idea is to stop pretending that one-size-fits-all `name` field - fits all causes of documentation, it does not, it does not have a mapping of semantics to the API of the function.

---

For example, imagine the signature:
```
dumpString :: forall m . IO.MonadIO m => (ByteString -> m ()) -> ByteString -> m ()
```

But this type signature puzzles because its type signature looks like `id (->)`: `dumpString :: forall m . IO.MonadIO m => (ByteString -> m ()) -> (ByteString -> m ())`.

Because type signature is not clear & puzzles - type signature needs explanation.
So minimal documentation would be attached to that.

We would also do not need in such cases to explain what particular `ByteString` or other repeating or alike types from several parameters passed we mean. Does "text" mean `ByteString` or in `IO` or `Text` or `String` - using ArgDocs it is harder to loose the reader because docs have 1:1 mapping & we think, writing, & explain in terms of the API.

Attaching documentation to type signature, ArgDocs - do not rot, because they are directly connected to the API.

For example flipping arguments in function - the main documentation block - would lose sequentiality - we either forget & sequentiality breaks & so the reader needs to infer words to args map - or dev would need to not forget to reread & rewrite the docs, which devs are famously forgetful about. Placing the main short API description into API signature - ties documentation to the realization & makes it be monitorable/trackable. In the project we frequently refresh type signatures for ourselves, but we do not reread the main documentation blocks.

And so on and so on. As you see - I have a lot of thoughts on this topic, which is why I went to fix the HLS API documentation interface to show them.

Anton-Latukha: As it is seen - generally programmers do not agree with me there, type signature documentation is thought about to be used so rarely - that HLS so far does not ship `show` of it.

Anton-Latukha: > I like this `traverse_ (yeald . str) [ "nix-archive-1", "(", .. ")" ]`

Yep, a nice example of meeting `Traverse` in the wild.

soulomoon: cool，I will add some rags doc when I was free today.

soulomoon: Sure, I am in no rush for this. lets wait a couple days.
Adding arg doc and making the project more readable is a good thing, 
especially when the project lacking manpower.

Anton-Latukha: > I never mind the longread, and having someone showing the direction and communicate is always a good thing

I also believe so, trying to stick to the dialectical method as the only scientific method of dialog possible, and so I also await good argumentation from people in a dialectic way to make a decision.

> especially when I am a newcomer to the this project, thank you for it

I trusted you from the first lines of code you send. I've seen that you have a good level of professionalism & that you wrote the code close to how I would write it. You also managed to implement & ship things in a complex codebase. It is possible to hold a productive dialog with you.

*Nota bene*: Mitchell Hashimoto fell in love with NixOS - which is a big thing, because HashiCorp is the most direct developer structure, he likes something - it gets implemented by the company in Enterprise masterpiece. So soon I expect much more movement around Nix or some isomorphic DLS & tide lifts all boats.

After you received an invitation to be a maintainer - you are one from the team.

soulomoon: Thank you for the invitation, let's make Hnix better.

hnix-store-core/src/System/Nix/Internal/Nar/Streamer.hs
Anton-Latukha: In such cases, you can just open another PR, which depends on this branch & mention it depends here & thankfully GitHub allows to mark the PR as a draft.

Having that one additional level of nesting is perfectly reasonable.

soulomoon: Sure the full version would be in draft #179

sorki:
Can you please rebase due to CI and maybe squash the minor fixes into respective commits?

soulomoon:
> Can you please rebase due to CI and maybe squash the minor fixes into respective commits?

Yeah, sure, I will adjust them later on.

sorki:
Somehow I've managed to miss the notification - there were some remnants of merges so I filtered this to only touch `-core` and they went away. CI broke meanwhile, so fixed that and rebased again and all looks good now.

PR #179, opened 2022-02-20
replace the filepath argument in addToStore with a more common type NarSource (The remote part)

soulomoon:
It depend on  #177


soulomoon:
The test would fail since the remote is using core in the hackage which is not updated to the newest core in master branch now, we should check again when a new core is updated in hacakge.

sorki:
> The test would fail since the remote is using core in the hackage which is not updated to the newest core in master branch now, we should check again when a new core is updated in hacakge.

`core-0.6` is now on Hackage. Added a commit that allows it for remote.

sorki:
Done via #190. Both released on Hackage as `0.6`. Thank you and sorry for the delay!

Issue #198, opened 2022-12-16
nar seralization hash mismatch


domenkozar:
For /nix/store/00mzv9k8gjw0pgzla4ysraq5kdsjbkb2-libffi-3.4.2, there's a mismatch in hashes comparing to what `nix-store --dump` and nar streaming would generate.

This hints there's bug somewhere.

domenkozar:
@sorki any ideas?

domenkozar:
To reproduce:

```
nix-store -r /nix/store/00mzv9k8gjw0pgzla4ysraq5kdsjbkb2-libffi-3.4.2
nix-store --dump /nix/store/00mzv9k8gjw0pgzla4ysraq5kdsjbkb2-libffi-3.4.2 > store
nix-shell -p "haskellPackages.ghcWithPackages (pkgs: [pkgs.hnix-store-core])"
❯ ghci
ghci> import System.Nix.Nar 
ghci> import System.IO
ghci> handle <- openFile "hnix" WriteMode
ghci> buildNarIO narEffectsIO "/nix/store/00mzv9k8gjw0pgzla4ysraq5kdsjbkb2-libffi-3.4.2" handle
diff -y <(xxd store) <(xxd hnix)
```

domenkozar:
Most likely we need to sort the files/directories list or something trivial like that.

Player-205:
Symlinks to the directories seems to be the reason of this issue. MRE:
```
$ cat default.nix
let
  pkgs = import <nixpkgs> {};
  in
derivation {
  system = "x86_64-linux";
  name = "test";
  builder = "${pkgs.bash}/bin/bash";
  args = [ ./builder.sh ];
  coreutils = pkgs.coreutils;
}
$ cat ./builder.sh
export PATH="$coreutils/bin"

mkdir $out
mkdir $out/1
echo 11 > $out/1/1

ln -s $out/1 $out/link1

$ nix-build  
this derivation will be built:
  /nix/store/67rnrqr2dwz9nqdv0awqrnvpv1gwjz2b-test.drv
building '/nix/store/67rnrqr2dwz9nqdv0awqrnvpv1gwjz2b-test.drv'...
/nix/store/bxgc56np7s0h35xl2dwmkcd65hxpxxqf-test
$ nix-store --dump /nix/store/bxgc56np7s0h35xl2dwmkcd65hxpxxqf-test > store
$ cabal repl hnix-store-core     
ghci> import System.Nix.Nar 
[...]

PR #201, opened 2022-12-27
[#198] Make nar implementation compatible with the original one

Player-205:
Problem: this implementation of nar made non-identical archives with the original implementation

Solution: doesn't treat the directory symlinks as they are directories

See https://github.com/haskell-nix/hnix-store/issues/198#issuecomment-1365911042 for the MRE


domenkozar:
Can you also make a test case?

domenkozar:
https://github.com/haskell-nix/hnix-store/blob/master/hnix-store-core/tests/NarFormat.hs

Player-205:
> Can you also make a test case?

Done

Issue #148, opened 2021-03-24
Reintroduce typelevel root paths, turn current `StorePath` into `SomeStorePath`


sorki:
While playing with `safe-money` library, I've realized that we don't need to commit on `TypeLits` and type level *only* `StorePath` and we can have both variants and couple of utility functions to turn one into other when/if needed.

I've quickly drafted the following (quite similar to `Dense` and `SomeDense` from `safe-money` + some questionable field duplication) and it looks plausible. After it compiled I went and checked prior implementation by @shlevy and it's quite similar, with the difference that this allows `SomeStorePaths` when we don't know/care about store root.

```haskell
data StorePath = StorePath
  { -- | The 160-bit hash digest reflecting the "address" of the name.
    -- Currently, this is a truncated SHA256 hash.
    storePathHash :: !(Digest StorePathHashAlgo)
  , -- | The (typically human readable) name of the path. For packages
    -- this is typically the package name and version (e.g.
    -- hello-1.2.3).
    storePathName :: !StorePathName
  }
  deriving (Eq, Ord)

newtype SymbolicStorePath (root :: Symbol) = SymbolicStorePath StorePath
  deriving (Eq, Ord)

data SomeStorePath = SomeStorePath {
    someStorePathHash :: !(Digest StorePathHashAlgo)
  , -- | The (typically human readable) name of the path. For packages
    -- this is typically the package name and version (e.g.
    -- hello-1.2.3).
    someStorePathName :: !StorePathName
  , -- | Root of the store
    someStorePathRoot :: !FilePath
  } deriving (Eq, Ord)

fromSomePath
  :: forall root
  .  KnownSymbol root
  => SomeStorePath
  -> Maybe (SymbolicStorePath root)
fromSomePath SomeStorePath{..} =
  if (someStorePathRoot == symbolVal @root Proxy)
  then Just $ SymbolicStorePath $ StorePath { storePathHash = someStorePathHash, storePathName = someStorePathName }
  else Nothing

withSomePath
  :: SomeStorePath
  -> (forall root. KnownSymbol root => SymbolicStorePath root -> a)
  -> a
```

Issue #62, opened 2020-03-21
Remove store type index to StorePath


Ericson2314:
Upstream C++ Nix switched to storing a pair of a hash and name, rather raw file path, for it's `StorePath` type. So far, this is just like us. However, there is no store type intrinsically associated with the store path in C++ Nix; if you wish to convert it to a concrete file path, you need to provide a store too so the missing information can be filled in. 

I think this is better than what we do. I understand we were using the type level trickery to mimick what old Nix did, but I think store-agnostic paths are better than store specific ones----all the moreso with things like content-addressed derivations making the `/nix/store` less and less important. (You can always relocate content-addressable outputs, so it doesn't matter if two stores agree.) Now that Nix made the switch to store-agnostic paths, we should too.

sorki:
This is addressed in #59.

PR #216, opened 2023-06-27
Do not store the store dir in the store path

Ericson2314:
I disagree with myself in https://github.com/haskell-nix/hnix-store/issues/62 (maybe I was confused on real vs virtual store dir, or maybe I was just being stupid!)

I agree with https://github.com/haskell-nix/hnix-store/issues/148 too, which is an issue to undo it.

As a first step, I present this, which merely make `StorePath` not store the store dir, with any type-level stuff TBD. Yes, this is not ideal, but I think it's a good start. It is how C++ Nix and TVix do it these day. It is essentially never safe to mix up store paths with different store dirs, and so this is an improvement under the "make illegal states unrepresentative front".

Some of the daemon code got uglier, but I have a WIP big overhaul of that I'd hope to finish up at some point. That undoes the slight nuisance of these "get store dirs" that crop up everywhere with this PR.

### DIFF
--- a/hnix-store-core/ChangeLog.md
+++ b/hnix-store-core/ChangeLog.md
@@ -1,5 +1,12 @@
 # ChangeLog
 
+## Unreleased 202y-mm-dd
+
+* Breaking:
+    * [(link)](https://github.com/haskell-nix/hnix-store/pull/216) `StorePath` no longer carries `storePathRoot` field and we
+      have a stand-alone `StoreDir` type instead to be used instead of `FilePath`
+      when store root directory is needed as a context.
+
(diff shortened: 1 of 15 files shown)


Ericson2314:
Thanks @sorki. Do you still have a working matrix account btw?

sorki:
> Thanks @sorki. Do you still have a working matrix account btw?

Not yet, but I was looking at dendrite last week and it looks pretty good, so I'll deploy it when I hit infra mood. We used to run our own synapse for a local NGO but it was scraped for "cost saving reasons" in favor of Slack. The real reason was that it was too transparent as anyone could join while their Slack is private and invite only :upside_down_face:

PR #218, opened 2023-11-08
Handle case conflicts on case-insensitive filesystems

### LINKED ISSUE #213: Nar: implement macOS case hack
https://github.com/NixOS/nix/blob/master/tests/functional/case-hack.sh

sandydoo:
Implements the "case hack" used by Nix to deal with case conflicts on hosts with case-insensitive filesystems, primarily macOS.

The case hack is enabled by default on macOS.

Resolves #213.

We'll have to add a NAR fixture to test this. You can't create the test case for this on the fly on a system that's case-insensitive.


sorki:
Looking good! Please rebase so it runs thru CI and add a changelog entry.

sandydoo:
Done! I filed this under "Additional". Could go under "Fixed" as well though.

sorki:
Very nice, thanks! Do you want a release of `core` or do you have some more stuff in pipeline?

sandydoo:
@sorki, that's it from me for now. A release would be great 🙂

PR #226, opened 2023-11-09
Fix NAR parsing failure when not enough input is available

### LINKED ISSUE #225: Nar parser fails with "not enough input" when parsing from a file handle
I've tracked it down to this `hGetSome`, which consistently returns fewer bytes than requested at the same exact position in the my sample NAR.

To my eye, this doesn't look quite right. Instead of throwing an error, we should be looping to fetch the remaining bytes. Or we replace the `hGetSome` with `hGet` and block.

https://github.com/haskell-nix/hnix-store/blob/f0dfc6d67015c271a9423ef9687aa57d4f097283/hnix-store-core/src/System/Nix/Internal/Nar/Parser.hs#L415-L429

#### Sample error with debugging info

```
parseSymlink: "/nix/store/m6dn1fp69fqgp0dc8pjapi5n8gf59b02-coreutils-full-9.3/bin/chmod"
Consuming 1 bytes
Expecting raw string
expectStr: ")"
Consuming 1 bytes
Expecting raw string
expectStr: ")"
Consuming 5 bytes
Expecting raw string
Consuming 1 bytes
Expecting raw string
expectStr: "("
Consuming 4 bytes
Expecting raw string
expectStr: "name"
Consuming 5 bytes
Expecting raw string
parseEntry: "choom"
Consuming 4 bytes
Expecting raw string
expectStr: "node"
Consuming 1 bytes
Expecting raw string
expectStr: "("
Consuming 4 bytes
Expecting raw string
expectStr: "type"
Consuming 7 bytes
Expecting raw string
Consuming 6 bytes
Expecting raw string
expectStr: "target"
Consuming 75 bytes
"/nix/store/yq3k4g6sw1q35ydv11csn"
Left "Exception while unpacking NAR file: user error (consume: Not enough bytes in handle. Wanted 75 got 32)"
```

sandydoo:
Replaces `hGetSome` with `hGet`, which should continue reading until the requested number of bytes have been read in. This avoids a potential failure mode where the requested bytes are not immediately available/buffered.

Resolves #225.

### DIFF
--- a/hnix-store-core/src/System/Nix/Internal/Nar/Parser.hs
+++ b/hnix-store-core/src/System/Nix/Internal/Nar/Parser.hs
@@ -421,7 +421,7 @@ consume
 consume 0 = pure ""
 consume n = do
   state0   <- State.get
-  newBytes <- IO.liftIO $ Bytes.hGetSome (handle state0) (max 0 n)
+  newBytes <- IO.liftIO $ Bytes.hGet (handle state0) (max 0 n)
   when (Bytes.length newBytes < n) $
     Fail.fail $
     "consume: Not enough bytes in handle. Wanted "


puffnfresh (approved):
Makes sense to me. Can't see a reason for `hGetSome` here.

sorki:
Thank you!

PR #231, opened 2023-11-15
remote: start transitioning from binary to cereal

sorki:
Strict by default, with round trip props.

Nix-like serialization primitives for now
via System.Nix.Store.Remote.Serialize.Prim


sorki:
I'll be glad If anyone wants to chime in and skim thru the commits as there's bunch more stuff coming. There are not many breaking changes to public interfaces, it's all in the spirit of tying the knots some more and getting both libs into better shape.

Ericson2314:
@sorki already knows, but just for others' benefit I just pushed https://github.com/obsidiansystems/hnix-store/tree/daemon-server-side which is a big overhaul of the remote protocol stuff. The plan is to integrate everything together one way or aonther.

sorki:
I'm going to merge this one as it allows to merge `hnix-store-db` in #70. The `StorePath.hs` is in disarray and will be cleaned-up shortly in follow-up PRs (after #234 is done).

Also the changelog would link to a huge PR which defies the point of linking to it and I'm too lazy to link to specific commits.

Issue #235, opened 2023-11-17
Split off `ReadonlyStore` into hnix-store-readonly


sorki:
Pre-req for #233. Should also be turned into `System.Nix.Store.ReadOnly`

Issue #233, opened 2023-11-17
Split off Nar into `hnix-store-nar`


sorki:
The thingie is pretty self-contained and doesn't even seem to depend on `-core`. The `ReadonlyStore` depends on it but that should also be a separate package it seems.

This would make dependencies cleaner and maintenance a bit easier.

CC @imalsogreg @domenkozar @sandydoo - I would add you all as maintainers on Hackage. Thoughts?

sorki:
Could also include [`hnix-store-benchmark`](https://github.com/domenkozar/hnix-store-core-benchmark) from https://github.com/haskell-nix/hnix-store/issues/26 and [`hnix-store-dump`](https://github.com/sandydoo/hnix-store-dump/) as executables.

sorki:
> Could also include [`hnix-store-benchmark`](https://github.com/domenkozar/hnix-store-core-benchmark) from #26 and [`hnix-store-dump`](https://github.com/sandydoo/hnix-store-dump/) as executables.

Since the thingie is a lib they should be guarded with flags like in [`-db`s cabal](https://github.com/haskell-nix/hnix-store/pull/70/files#diff-b7b645e18ed85cbf7c141011b6de7a3475630df1b5387fe197f2b20aaf54ef05R19-R29) + [buildable](https://github.com/haskell-nix/hnix-store/pull/70/files#diff-b7b645e18ed85cbf7c141011b6de7a3475630df1b5387fe197f2b20aaf54ef05R92-R116) so they won't propagate to users PATH by default but enabled in dev env via `cabal.project`

sorki:
When this lands on Hackage, the migration path for `cachix` is simple:
- add lower bound to `hnix-store-core` to require `>= 0.8`
- add `hnix-store-nar`

sorki:
Keeping open as the exe and benchmark would be cool! I think Domen and Sander wouldn't mind if this is picked from their repos and added here with `git commit --author=...` attribution

PR #241, opened 2023-11-20
Split core into multiple packages

sorki:
- First one is `-tests`
- `-nar` #233
- `-readonly` #235
will follow


Ericson2314 (approved):
I didn't read this all by any means, but from talking with @sorki and glancing it over it seems to be the right idea

sorki:
Need to practice small-er PRs so I'll open another one for NAR/readonly.

PR #243, opened 2023-11-21
remote: fix queryPathInfoUncached narHash decoding

squalus:
nix-daemon encodes the narHash as Base16, not NixBase32: https://github.com/NixOS/nix/blob/a6b315ae/src/libstore/worker-protocol.cc#L180

Change the narHash decoding step to use Base16.

Add a test that fails on the previous code with the following error:

    uncaught exception: ErrorCall
    Invalid NixBase32 string
    CallStack (from HasCallStack):

Tested with nix 2.13.6.

### DIFF
--- a/hnix-store-remote/src/System/Nix/Store/Remote.hs
+++ b/hnix-store-remote/src/System/Nix/Store/Remote.hs
@@ -46,7 +46,7 @@ import           System.Nix.Build               ( BuildMode
                                                 , BuildResult
                                                 )
 import           System.Nix.Hash                ( NamedAlgo(..)
-                                                , BaseEncoding(NixBase32)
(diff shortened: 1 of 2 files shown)


squalus:
I would have expected the previous test to fail in the same way. I'm not sure what's going on there.

sorki:
Probably just wasn't evaluated due to laziness.

Take a look at a branch by @obsidiansystems, many of these are fixed there and I'm going to start integrating that today (pre-reqs for core are already done)  https://github.com/haskell-nix/hnix-store/compare/master...obsidiansystems:hnix-store:daemon-server-side#diff-757c34c188c77626929a2bd56ab6716f7724cebddae7b29f4cd8dda7b6399b8b

Issue #240, opened 2023-11-20
Steal signature tests from tvix-compat


sorki:
https://cl.tvl.fyi/c/depot/+/10081/1/tvix/nix-compat/src/narinfo/signature.rs

With proper attribution to @flokli!

This might allow us to deal with saltine vs cryptonite/crypton conundrum.

flokli:
The credit also goes to @zimbatm, the fixture was produced as part of https://github.com/nix-community/go-nix/pull/93.

zimbatm:
This kind of stealing is good!

PR #244, opened 2023-11-21
remote: fix queryPathInfoUncached missing content address decoding

squalus:
queryPathInfoUncached fails with a "not enough input" message when decoding Metadata that is missing a content address.

nix-daemon encodes a missing content address as an empty string: https://github.com/NixOS/nix/blob/a6b315ae8/src/libstore/content-address.cc#L139

Change the decoding to return Nothing if the caString is empty.

I manually tested this with the following program:

```haskell
{-# LANGUAGE OverloadedStrings #-}

module Main

where

import System.Nix.Store.Remote
import System.Nix.StorePath
import Control.Monad (void)
import Control.Monad.IO.Class (liftIO)
import Data.Default.Class

main :: IO()
main = do
  void $ runStore $ do
    let Right sp = parsePath def "/nix/store/sbldylj3clbkc0aqvjjzfa6slp4zdvlj-hello-2.12.1"
    q <- queryPathInfoUncached sp
    liftIO $ print q
```
The store path came from `nix build nixpkgs#hello`. Tested with nix 2.13.6.

### DIFF
--- a/hnix-store-remote/src/System/Nix/Store/Remote.hs
+++ b/hnix-store-remote/src/System/Nix/Store/Remote.hs
@@ -34,6 +34,7 @@ import Data.Dependent.Sum (DSum((:=>)))
 import Data.HashSet (HashSet)
 import Data.Map (Map)
 import Data.Text (Text)
+import qualified Data.Text
 import qualified Control.Monad
 import qualified Data.Attoparsec.Text
 import qualified Data.Text.Encoding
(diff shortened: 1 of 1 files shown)


sorki:
Also done in `daemon-server-side` https://github.com/haskell-nix/hnix-store/pull/243#issuecomment-1822150182

May I ask you to take a look at #246 and #240? That would help me a lot and I'll deal with the `-remote` integration part meanwhile so you don't have to duplicate the work/fixes done there.

sorki:
Or any of the recent issues marked as good first issue, which I won't do right now so I can focus on unblocking `-remote`.

PR #255, opened 2023-11-23
cereal remote, server side integration

sorki:
- binary -> cereal
- start of the https://github.com/obsidiansystems/hnix-store/tree/daemon-server-side integration


sorki:
@squalus `Arbitrary Signature` instance was quite helpful to allow `Arbitrary (Metadata StorePath)`, thanks for adding that!

sorki:
Serializer done, error handling done, GC done, `build-derivation` executable is broken but easy to fix but if I move `ourProtoVersion` some tests start to fail, will resolve in next PR. 

Progressing slowly but the gradual refactoring without breaking tests is the way to go, the rest should land swiftly.

Issue #253, opened 2023-11-22
`-nar` testsuite uses binary


sorki:
but the library uses `cereal`. Switch testsuite to `cereal` as well.

sorki:
Done in #258

PR #258, opened 2023-11-25
Use `cereal` instead of `binary` in -nar test suite

vaibhavsagar:
Addresses https://github.com/haskell-nix/hnix-store/issues/253.


Ericson2314 (approved):

sorki:
Added a commit removing leftover comment. 

Thank you!

Issue #252, opened 2023-11-22
Port e5f1c7d to `hnix-store-nar`


sorki:
e5f1c7d9bcf618c3d1351fb006699285402f42d5 is part of https://github.com/haskell-nix/hnix-store/pull/41 and the only thing worth preserving from the PR.

PR #259, opened 2023-11-25
Replace permissions with `narIsExec` effect

vaibhavsagar:
Addresses https://github.com/haskell-nix/hnix-store/issues/252.


sorki:
Nice! Can you also add `Co-Authored-By: Shea Levy <shea@shealevy.com>` to the commit message so the original author attribution is there. Forgot to mention that in the issue.

vaibhavsagar:
Sure, updated.

PR #260, opened 2023-11-26
Set executable permissions once instead of multiple times

vaibhavsagar:
In https://github.com/haskell-nix/hnix-store/pull/259 I made a mistake and updated `streamStringOutIO` to set the executable permission once per chunk of the input. This fixes that and moves that logic to the end so it is only done once.


sorki:
Nice catch, bonus points if you could add a test :smiley_cat:

vaibhavsagar:
I'm not sure how to add a good test for this, the previous (broken) implementation has the same overall behaviour but it does the same thing multiple times which I think should be less performant.

sorki:
No problem, I didn't really gave this much thought as you know better at this point, it could always be added later if needed or you stumble on a solution. Thanks!

PR #264, opened 2023-11-26
core: add fingerprints

squalus:


sorki:
Looks good, clearly useful, comes with a test, I'm happy :smile_cat: Only fixed a `-- ^` to `-- |` for function haddocks.

PR #276, opened 2023-12-08
Some more server

sorki:


hnix-store-core/src/System/Nix/DerivedPath.hs
```diff
@@ -25,12 +25,17 @@ import qualified System.Nix.StorePath
 
 data OutputsSpec =
     OutputsSpec_All
+  -- ^ Wildcard spec (!*) meaning all outputs
```
Ericson2314, line 28:
```suggestion
  -- ^ Wildcard spec (^*) meaning all outputs
```
the syntax in the worker protocol is somewhat deprecated

sorki: Fixed, thanks!

sorki:
Thanks to @ObsidianSystems for promting all this and the code of course. Especially to @Ericson2314 for being my rubber duck and many good pointers but also @squalus for fingerprints and @layus for typeclass/tagless PoC at #72.

More to come!

PR #277, opened 2023-12-12
remote: add AddToStoreNar operation

### LINKED ISSUE #265: Implement AddStoreToNar operation
My use case is that I have a bunch of NARs and narinfos and I'd like to use hnix-store-remote to add the NARs to the Nix Store. To do this, hnix-store-remote needs to support the addToStoreNar operation. It also must not read the entire NAR into memory, since the NARs can be quite large.

I have some code that does this with `conduit`. I imagine using `ByteString.Lazy` would be a better fit for this project though.

addToStoreNar support was removed in #75

squalus:
Some open items:

- What type should be used for the reply? Nothing is sent over the wire. `()` is already taken by `opSuccess` which reads a boolean. Maybe `opSuccess` should use a new reply type, and this one should use `()`
- The client chooses the chunk size. What should that be?
- Better errors for the `RemoteStoreError_Fixme` usages
- Make an Arbitrary instance for `SampleNar`

Closes #265

### DIFF
--- a/hnix-store-remote/src/System/Nix/Store/Remote/Client/Core.hs
+++ b/hnix-store-remote/src/System/Nix/Store/Remote/Client/Core.hs
@@ -7,9 +7,13 @@ module System.Nix.Store.Remote.Client.Core
 import Control.Monad (unless, when)
 import Control.Monad.Except (throwError)
 import Control.Monad.IO.Class (MonadIO, liftIO)
+import Data.ByteString (ByteString)
 import Data.DList (DList)
 import Data.Some (Some(Some))
+import Data.Word (Word64)
+import Network.Socket (Socket)
 import System.Nix.Nar (NarSource)
+import System.Nix.StorePath.Metadata (Metadata(..))
 import System.Nix.Store.Remote.Logger (processOutput)
 import System.Nix.Store.Remote.MonadStore
   ( MonadRemoteStore(..)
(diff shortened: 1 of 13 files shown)


sorki:
Very nice (especially the framing!) and sorry for the conflicts! Wanted to merge earlier but didn't get to it sooner.

The `CheckSignatures` is also in core (I've realized it is actually `Check`Signatures as well only couple days ago) - but don't worry about it, I'll deal with it.

You're right about `opSuccess`, most ops simply return `True` and indicate errors via logger so it seemed appropriate to just use `()` but this one makes it weird, I agree about the newtype.

No idea about the chunk size tbh.

hnix-store-remote/src/System/Nix/Store/Remote/Client/Core.hs
```diff
@@ -75,14 +81,57 @@ doReq = \case
...
+      AddToStoreNar _ meta _ _ -> do
+        let narBytes = maybe 0 id $ metadataNarBytes meta
+        maybeDataSource <- takeDataSource
+        soc <- getStoreSocket
+        case maybeDataSource of
+          Nothing ->
+            if narBytes == 0 then writeFramedSource (const (pure Nothing)) soc 0
+            else throwError RemoteStoreError_NoDataSourceProvided
+          Just dataSource -> do
+            writeFramedSource dataSource soc narBytes
```
sorki, line 95: Now I kinda wonder what's the `Logger_Read` code path for.

squalus:
- Replaced the existing `()` reply type with a new `SuccessCodeReply` type
- Added a `NoReply` type for this case
- Added error types
- Resolved conflicts

hnix-store-remote/src/System/Nix/Store/Remote/Client/Core.hs
squalus: Yeah, I don't know if it's still used in the newer protocol versions.

sorki:
Nicely done, thank you!

sorki:
Btw I'm pretty much done with most large-scale changes/refactors so things should settle down now, feel free to tackle what you see fit.

PR #279, opened 2023-12-14
remote: add NarFromPath client

### LINKED ISSUE #266: Implement NarFromPath operation
This would allow exporting NAR files from a remote Nix store. This should not read the entire NAR into memory, since the NARs can be large.

squalus:
Adds a client for the NarFromPath operation along with a small test.

The NAR size is not provided by the server. The user needs to provide it to the NarFromPath call. It determines how many bytes get read from the server. If the client provides the wrong byte size, the call will fail in a bad way, like hanging open. I don't think there's any way to avoid this.

The size isn't part of the wire protocol so it can't go in the StoreRequest type (roundtrip tests wouldn't work). I elected to put it in RemoteStoreState alongside the sink. This seems unpleasant. There might be a better way to arrange this.

Closes #266

### DIFF
--- a/hnix-store-remote/src/System/Nix/Store/Remote/Client.hs
+++ b/hnix-store-remote/src/System/Nix/Store/Remote/Client.hs
@@ -11,6 +11,7 @@ module System.Nix.Store.Remote.Client
   , ensurePath
   , findRoots
   , isValidPath
+  , narFromPath
   , queryValidPaths
   , queryAllValidPaths
   , querySubstitutablePaths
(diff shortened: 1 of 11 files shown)


sorki:
> The size isn't part of the wire protocol so it can't go in the StoreRequest type (roundtrip tests wouldn't work). I elected to put it in RemoteStoreState alongside the sink. This seems unpleasant. There might be a better way to arrange this.

No problem with that, most of the `MonadRemoteStore` typeclass funs are sort-of internal and I doubt someone will need to override these except for `appendLogs` which would be helpful to change without also having to boilerplate the rest. (I don't know a good solution to that yet).

hnix-store-remote/src/System/Nix/Store/Remote/Client.hs
```diff
@@ -181,6 +182,18 @@ isValidPath
...
+narFromPath
+  :: MonadRemoteStore m
+  => StorePath -- ^ Path to generate a NAR for
+  -> Word64 -- ^ Byte length of NAR
+  -> (ByteString -> IO()) -- ^ Data sink where NAR bytes will be written
+  -> m ()
+narFromPath path narSize sink = do
+  setDataSink sink
+  setDataSinkSize narSize
+  void $ doReq (NarFromPath path)
```
sorki, line 195: If you always need to do `queryPathInfo` to figure out the size before, you might just add it here and possibly return its result as well.

sorki:
Thanks!

PR #286, opened 2024-10-04
nar: fix ordering of case-hacked paths on macOS

sandydoo:
When undo-ing the case hack on macOS, we need to restore the original order of paths when traversing directories. This entails undo-ing the case hack for the entire listed directory and then sorting.


domenkozar:
@sorki can you make a release :pray:

sorki:
Yep, released as `hnix-store-nar-0.1.1.0` via https://github.com/haskell-nix/hnix-store/pull/287

Thanks!

PR #288, opened 2024-10-31
Content address method cleanup

Ericson2314:
Doing some refactors I did in upstream Nix; each commit here links the commits there.


Ericson2314:
Now contains #291; very nice!

hnix-store-readonly/src/System/Nix/Store/ReadOnly.hs
```diff
@@ -27,91 +31,110 @@ import qualified System.Nix.Hash
...
-  if recursive == FileIngestionMethod_FileRecursive
-     && (algoName @hashAlgo) == "sha256"
-  then makeStorePath storeDir "source" h
-  else makeStorePath storeDir "output:out" h'
+makeFixedOutputPath storeDir method digest@(hashAlgo :=> h) refs =
+  case method of
+    ContentAddressMethod_Text ->
+      case hashAlgo of
+        HashAlgo_SHA256 -> makeTextPath storeDir h $ references_others refs
+        _ -> error "unsupported" -- TODO do better; maybe we'll just remove this restriction too?
```
Ericson2314, line 109: Maybe should make this better :)

hnix-store-core/src/System/Nix/Store/Types.hs
```diff
@@ -1,17 +1,11 @@
+-- | TODO rename module
```
sorki, line 1: Split into `RepairMode.hs` and `PathFilter.hs`? Not sure if worth just for these two.

sorki (approved):

PR #289, opened 2024-10-31
Inline `nix-derivation` (and PRs), extend

Ericson2314:
- Move the derivation types to core, cut down parameters somewhat
- Rename (the rest of) nix-derivation to hnix-store-aterm, since I want to de-center ATerm as "the canonical form", and put it on the same level as the JSON.

~~Depends on #288~~

Things remaining:

- [x] Test failures (`hnix-store-remote`)
   - https://github.com/obsidiansystems/hnix-store/pull/2 PR description has some discussion of what remains
- [x] JSON store paths should never include store dir as of unreleased Nix (https://github.com/NixOS/nix/pull/13980, issue https://github.com/NixOS/nix/issues/13570)
    - https://github.com/haskell-nix/hnix-store/pull/289/files#r2363911656 comment of example problem in this PR

PR #290, opened 2024-11-01
wip adapt to new Derivation

sorki:
Not quite there yet as we're missing some parsers. `-remote` not fully adapted either.


Ericson2314:
I think we should just go with https://github.com/haskell-nix/hnix-store/pull/289 because I had to change things above and beyond what your PR was doing.

sorki:
Sounds good, your changes to parsers seems more elegant anyway, don't quite remember why I've had to dupe that much, possibly because I didn't want to change the `nix-derivation` ones.

PR #305, opened 2026-02-11
Cryptonite ghost

yaitskov:
@sorki


sorki:
Solved in https://github.com/haskell-nix/hnix-store/pull/307

PR #316, opened 2026-06-04
hnix-store-json: Include version in Derivation

artemist:
The `nix derivation add` command now requires a version in the imported json and `nix derivation show` includes a version, both version 4.


Ericson2314 (approved):

Ericson2314:
Note for later: we should parse the version too and make sure it is one that we expet

